PR #66, opened 2025-11-03
indexer operator

james1314520:
a fp8_indexer_lighting operator


examples/fp8_indexer_lighting.py
CeleNewYear: I found that the precision is incorrect in my test. Please review and modify the code before submitting it again.

PR #109, opened 2025-12-02
[Feature] support Persistent

mcl910:


examples/gemm/example_gemm_intrinsic_persistent.py
```diff
@@ -0,0 +1,141 @@
...
+            })
+
+            A_L0 = T.alloc_L0A((S2, block_M, block_K), dtype)
+            B_L0 = T.alloc_L0B((S2, block_K, block_N), dtype)
+            C_L0 = T.alloc_L0C((block_M, block_N), accum_dtype)
+
+            with T.Scope("C"):
+                init_flag()
+
+                for i in T.serial(T.ceildiv(m_num * n_num, core_num)):
```
xwhzz: Why keep this line?

mcl910: Thank you, I have noticed this issue. Indeed, this line will cause an additional layer of loops in the production code, which is the reason for the significant decrease in performance. I will remove this line。

PR #129, opened 2025-12-11
Support for memory operations in develop mode.

XMU-qcj:
Supports end-to-end testing for Vector and Cube kernels. Test cases can be found in examples/matmul_develop.py and vec_add_1d_develop.py. End-to-end testing is not yet supported for mix kernels, only generation with bishengir is supported.

### DIFF
--- a/examples/matmul_develop.py
+++ b/examples/matmul_develop.py
@@ -1,0 +1,66 @@
+# Copyright (c) Huawei Technologies Co., Ltd. 2025.
+import torch
+import tilelang
+import tilelang.language as T
+
+torch.npu.set_device(0)
+tilelang.cache.clear_cache()
+
+M = 1024
+N = 512
+K = 2048
+
+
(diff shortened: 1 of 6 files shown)


baberrehman:
We need to create a new branch for developer mode based on expert mode.

baberrehman:
Please add a minimal example of how/which memory ops changed in developer and expert mode.

XMU-qcj:
> We need to create a new branch for developer mode based on expert mode.

The current operations on memory ops are fully compatible with the subsequent expert mode, and the two can also coexist.

XMU-qcj:
> Please add a minimal example of how/which memory ops changed in developer and expert mode.

Yes, my vec_add_1d_develop.py is a mix of memory operations from develop and expert, while matmul_develop.py is purely develop. I submitted both cases, which should clear up your confusion.

PR #133, opened 2025-12-12
[Bugfix] Fix the redundant cast & copy issue in examples/gemv

liggest:


benyang0506:
Why is the file named gemv_v and not gemv?

liggest:
> Why is the file named gemv_v and not gemv?

It's a implementation of gemv using vector core.
I assume that there can be an implementation using the cube core, but so far I haven't managed to realize that.

PR #207, opened 2026-01-04
[Enhancement] enable passing basic dtype var as args

BUPTAntiRiver:
Previously, codegen considers all params as handle. Added support for non-handle cases.


BUPTAntiRiver:
Fixes #149

PR #210, opened 2026-01-04
[AscendNPU-IR][Expert]: Support reduce and reduce_{max,min,sum} ops, and add clear=false handling

lzp1021:
1. Add support for reduce and reduce_{max,min,sum} operations.
2. Extend support to handle cases where clear is set to false.


tilelang/language/reduce.py
```diff
@@ -26,16 +27,19 @@ def reduce(buffer: tir.Buffer, out: tir.Buffer, reduce_type: str, dim: int, clea
...
-        buffer,
-        out,
+        tir.op.Op.get("tl.npuir_reduce"),
+        src,
+        dst,
+        str(dim),
         reduce_type,
-        dim,
-        clear,
+        clear
```
CeleNewYear: Why modify this part? The changes in customize_npuir.py should already be sufficient to support the reduce functionality.

PR #223, opened 2026-01-06
[AscendNPU-IR][Expert]: Support ClampOp

WCY2000:


testing/npuir/test_clamp
```diff
@@ -0,0 +1,101 @@
...
+        with T.Kernel(BLOCK_SIZE, is_npu=True) as (cid, _):
+            # Allocate UB (Unified Buffer) memory
+            src_ub = T.alloc_ub((M, N), dtype)
+            dst_ub = T.alloc_ub((M, N), accum_dtype)
+
+            # Copy from GM (Global Memory) to UB
+            T.copy(src, src_ub)
+
+            # Clamp operation
+            T.clamp(src_ub, dst_ub, CLAMP_MIN, CLAMP_MAX)
```
CeleNewYear, line 41: Support for two types of scenarios is required.

One type is the demo you provided, where the clamp upper and lower limits are constants.
Another type is when the upper and lower bounds are vectors; please support this scenario as well.

CeleNewYear (changes requested):

PR #226, opened 2026-01-07
[AscendNPU-IR][Developer] support brc op

yyyy999:
Support brc in develop mode

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -1074,6 +1074,11 @@ void CodeGenTileLangNPUIRDEV::VselectCodegen(const CallNode *op) {
   selOp->setAttr("broadcast", builder.getDenseI64ArrayAttr(broadcastDim));
 }
 
+/// Generate hivm.hir.vbrc for tl.npuir_brc.
+/// before:
+///    T.npuir_brc(A, B)
+/// after:
+///    %.* = hivm.hir.vbrc ins(A) outs(B) -> tensor<>
 void CodeGenTileLangNPUIRDEV::VbrcCodegen(const CallNode *op) {
   tvm::tl::NpuirBrc npuirop(op->args, this->vmap);
   mlir::Value src;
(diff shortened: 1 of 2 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -1124,16 +1132,24 @@ void CodeGenTileLangNPUIRDEV::VcastCodegen(const CallNode *op) {
...
       mlir::hivm::ReduceOpAttr::get(&context, NPUIR_STR_REDUCEOP[reduce_mode]);
-  builder.create<mlir::hivm::VReduceOp>(
-      builder.getUnknownLoc(), TypeRange{}, src, dst, mode,
+  mlir::Type dst_type = dst.getType();
+  mlir::TypeRange result_tensors(&dst_type, 1);
+  auto newCastOp = builder.create<mlir::hivm::VReduceOp>(
+      builder.getUnknownLoc(), result_tensors, src, dst, mode,
       builder.getDenseI64ArrayAttr(npuirop.reduce_dims));
+  SetVarValue(npuirop.dst, newCastOp->getResult(0));
 }
```
CeleNewYear, line 1176: Another PR has solved this, No need change

src/target/codegen_npuir_dev.cc
```diff
@@ -1074,6 +1074,11 @@ void CodeGenTileLangNPUIRDEV::VselectCodegen(const CallNode *op) {
   selOp->setAttr("broadcast", builder.getDenseI64ArrayAttr(broadcastDim));
 }
 
+/// Generate hivm.hir.vbrc for tl.npuir_brc.
```
CeleNewYear, line 1077: Please add a test case for tl.npuir_brc

CeleNewYear (approved):

src/target/codegen_npuir_dev.cc
```diff
@@ -1086,20 +1091,23 @@ void CodeGenTileLangNPUIRDEV::VbrcCodegen(const CallNode *op) {
       src = MakeValue(npuirop.in);
     }
   } else {
-    src = GenSubviewFromRegion(npuirop.src, npuirop.src_range);
-    auto srcMemref = llvm::dyn_cast<TypedValue<MemRefType>>(src);
-    inBufferShape = srcMemref.getType().getShape();
+    src = GetVarValue(npuirop.src);
```
baberrehman, line 1094: Do we not support subview on tensors?

baberrehman: And why do we need to call GetVarValue here?

src/target/codegen_npuir_dev.cc
```diff
@@ -1086,20 +1091,23 @@ void CodeGenTileLangNPUIRDEV::VbrcCodegen(const CallNode *op) {
...
-    auto outBufferShape = outMemref.getType().getShape();
+    auto outTensor = llvm::dyn_cast<TypedValue<RankedTensorType>>(dst);
+    auto outBufferShape = outTensor.getType().getShape();
     auto broadcastDim = getBroadcastDim(npuirop.src->shape, npuirop.dst->shape);
     broadcastDimAttr = builder.getDenseI64ArrayAttr(broadcastDim);
   }
-  builder.create<mlir::hivm::VBrcOp>(builder.getUnknownLoc(), TypeRange{},
+  mlir::Type dst_type = dst.getType();
+  mlir::TypeRange result_tensors(&dst_type, 1);
+  auto newCastOp = builder.create<mlir::hivm::VBrcOp>(builder.getUnknownLoc(), result_tensors,
```
baberrehman, line 1108: Why do we need to pass `result_tensors`, can we not just pass `TypeRange{}` and let the op infer the output type?

baberrehman (changes requested):

src/target/codegen_npuir_dev.cc
yyyy999: > Do we not support subview on tensors?
Yes, GenSubviewFromRegion only support on memref

> And why do we need to call GetVarValue here?
For tensor support, use GetVarValue we can get the tensor value

src/target/codegen_npuir_dev.cc
yyyy999: It cannot inter automatically, so we need to pass the result_tensor

baberrehman: Interesting, but the provided test runs successfully with `mlir::TypeRange{}`.

baberrehman: It's because the type inference doesn't work for tensor (it has not been implemented for tensor types).

src/target/codegen_npuir_dev.cc
baberrehman: Can we use `MakeValue` instead of `GetVarValue`?

baberrehman (changes requested):

src/target/codegen_npuir_dev.cc
baberrehman: I am creating a separate issue to track this: https://github.com/tile-ai/tilelang-ascend/issues/250

baberrehman (approved):

baberrehman (approved):

baberrehman (approved):

PR #235, opened 2026-01-08
[AscendNPU-IR] [Developer]: handle different src and dst shape for AscendCopy op

taowenqi:


src/target/codegen_npuir_dev.cc
```diff
@@ -997,9 +997,84 @@ void CodeGenTileLangNPUIRDEV::AscendCopyCodegen(const CallNode *op) {
...
+        for (int i = 0; i < dst_memref_type.getRank(); ++i) {
+          offsets.push_back(builder.getIndexAttr(0));
+        }
+        
+        // Set sizes to target shape
+        for (int64_t dim : dst_memref_type.getShape()) {
+          if (mlir::ShapedType::isDynamic(dim)) {
+          // For dynamic dimensions, we need to obtain them from elsewhere, simplified here
+          // In actual code, you may need to get dynamic dimension values from operands
+            auto dim_val = builder.create<mlir::arith::ConstantIndexOp>(loc, 1);
```
CeleNewYear: Please try setting it to -1. I remember dynamic dim automatically matches when the value is -1.

src/target/codegen_npuir_dev.cc
```diff
@@ -997,9 +997,84 @@ void CodeGenTileLangNPUIRDEV::AscendCopyCodegen(const CallNode *op) {
...
     dst_value = convertTensorToMemref(dst_value);
-    builder.create<mlir::memref::CopyOp>(
-        builder.getUnknownLoc(), TypeRange{}, src_value, dst_value
-    );
+      
+    auto src_memref_type = src_value.getType().cast<mlir::MemRefType>();
+    auto dst_memref_type = dst_value.getType().cast<mlir::MemRefType>();
+
+    // If shapes are inconsistent but total number of elements is the same, try to reshape the source
+    if (src_memref_type.getShape() != dst_memref_type.getShape() &&
```
baberrehman, line 1106: Do we really need to handle different shapes in `src` and `dst`? Shouldn't this be an error in compilation?

baberrehman (changes requested):

baberrehman:
Please also update commit message as: [AscendNPU-IR] [Developer]: handle different src and dst shape for AscendCopy op

src/target/codegen_npuir_dev.cc
taowenqi: > Do we really need to handle different shapes in `src` and `dst`? Shouldn't this be an error in compilation?


This actually happens when:
            A_BUF = T.alloc_shared((block_N), dtype)

            for i in T.serial(M):
                T.copy(A[i, :], A_BUF) 
                T.copy(A_BUF, B[i, :])
we'd like to allocate memory like [1, block_N], but to keep consistent with tilelang offical expression, we use T.alloc_shared(block_N), which is not consistent with the actual memory size. Different shapes mainly points to this situation.

baberrehman (approved):

PR #237, opened 2026-01-08
[AscendNPU-IR][Developer]: Support copy op from tensor to tensor

mmyy1217:
1. Add support for copy op from tensor to tensor using insert_slice
2. Add simple copy example
3. Refract cast op insertion in copy as createCastIfTypeMismatch

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -964,6 +964,7 @@ void CodeGenTileLangNPUIRDEV::AscendCopyCodegen(const CallNode *op) {
     }
     return value;
   };
+  // Cast if src and dst type mismatch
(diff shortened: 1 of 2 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -959,35 +959,83 @@ void CodeGenTileLangNPUIRDEV::AscendCopyCodegen(const CallNode *op) {
     }
     return value;
   };
+  auto createCastIfTypeMismatch = [&](mlir::Value src_value, mlir::Value dst_value) -> mlir::Value {
```
CeleNewYear, line 968: Add the corresponding comment.

mmyy1217: done

PR #243, opened 2026-01-10
[AscendNPU-IR][Developer] Support add/Sub/mul/div in T.Pipelined

CeleNewYear:
Added support for vector add and vector mul in the T.Pipelined OP

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -2767,20 +2767,47 @@ void CodeGenTileLangNPUIRDEV::VisitStmt_(const DeclBufferNode *op) {
   VisitStmt(op->body);
 }
 
-void CodeGenTileLangNPUIRDEV::LoopCarriedVarCollector::VisitExpr_(const tir::CallNode* call) {
+void CodeGenTileLangNPUIRDEV::LoopCarriedVarCollector::VisitExpr_(
+    const tir::CallNode *call) {
+  auto check_var = [&](const tir::VarNode *var_node) {
+    if (var_node && outer_->GetVarValue(var_node) != mlir::Value{}) {
+      loop_carried_vars_.insert(var_node);
+    }
+  };
+  auto process_call_arg = [&](int arg_index) {
+    auto &arg = call->args[arg_index];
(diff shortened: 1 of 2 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -2400,19 +2400,34 @@ void CodeGenTileLangNPUIRDEV::VisitStmt_(const DeclBufferNode *op) {
...
-      auto check_var = [&](const tir::VarNode* var_node) {
-      if (var_node && outer_->GetVarValue(var_node) != mlir::Value{}) {
-        loop_carried_vars_.insert(var_node);
-      }
-    };
-    
     check_var(npuirop.src0->data.get());
     check_var(npuirop.src1->data.get());
     check_var(npuirop.dst->data.get());
+  } else if (call->op.same_as(Op::Get("tl.npuir_add")) ||
```
CeleNewYear, line 2794: Please Add test cased

src/target/codegen_npuir_dev.cc
```diff
@@ -2399,20 +2399,47 @@ void CodeGenTileLangNPUIRDEV::VisitStmt_(const DeclBufferNode *op) {
   VisitStmt(op->body);
 }
 
-void CodeGenTileLangNPUIRDEV::LoopCarriedVarCollector::VisitExpr_(const tir::CallNode* call) {
+void CodeGenTileLangNPUIRDEV::LoopCarriedVarCollector::VisitExpr_(
+    const tir::CallNode *call) {
+  auto check_var = [&](const tir::VarNode *var_node) {
+    if (var_node && outer_->GetVarValue(var_node) != mlir::Value{}) {
```
baberrehman, line 2773: What if `var_node` already exists in `loop_carried_vars_`? Do we need this check?

baberrehman (changes requested):

src/target/codegen_npuir_dev.cc
CeleNewYear: Do not, because loop_carried_vars_ is a set

baberrehman (approved):

Issue #241, opened 2026-01-09
Request for T.Parallel


LLMZhangYC:
Future Feature Requests for Parallel：
1. Automic copy from Ub to GM
    `for i,j in T.Parallel(row, col):
          B_GM[i, j] = A_UB[i, j] + bias`
2. Support for element offset
   `for i,j in T.Parallel(row, col):
          B[i, j] = A[i + h, j] + bias`
3. Support Mismatched Operand-Result Dimensions
    `for i,j in T.Parallel(row, col):
          B[i, j] = A[i] + bias`

z00520135:
1、Automic copy from Ub to GM
for i,j in T.Parallel(row, col): B_GM[i, j] = A_UB[i, j] + b 
answer:We already support the scenario of ub ->gm and will support gm->ub/gm->l01/loc->gm in this month
2、Support for element offset
for i,j in T.Parallel(row, col): B[i, j] = A[i + h, j] + bias
answer:T.Parallel can handle cases where elements has offset like: a[i, j + scalar],process util scalar of each tile,but if you want treating him as an element requires the next generation chip SIMT mode
3、Support Mismatched Operand-Result Dimensions
for i,j in T.Parallel(row, col): B[i, j] = A[i] + bias
answer:we will support in this month

PR #258, opened 2026-01-13
Support element offset in T.Parallel

BUPTAntiRiver:
Also fix two typos in example


BUPTAntiRiver:
#241 request number 2

PR #261, opened 2026-01-14
Suppot T.Parallel element offset

BUPTAntiRiver:
Now T.Parallel can handle cases where elements has offset like: a[i, j + scalar], fix vectorization according to Parallel vars rather than just infer from buffer shape, which is more flexible.
Also fix two little bugs in example.


BUPTAntiRiver:
Fix #241 element offset request

PR #264, opened 2026-01-14
[AscendNPU-IR][Developer][Expert]: Support AscendCopy op for different src and dst dynamic shape

taowenqi:
These changes have been rebased and tested on NPU.

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -1086,8 +1086,7 @@ void CodeGenTileLangNPUIRDEV::SmartMemRefCopy(mlir::Value src, mlir::Value dst) 
   }
 
   // 2. Try Reinterpret Copy if shape not match but numElements match
-  if (src_type.getNumElements() == dst_type.getNumElements() && 
-      src_type.getElementType() == dst_type.getElementType()) {
+  if (src_type.getNumElements() == dst_type.getNumElements()) {
       
     // safety check: stride hack only when elementTypes match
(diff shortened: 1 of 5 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -1087,62 +1087,125 @@ void CodeGenTileLangNPUIRDEV::SmartMemRefCopy(mlir::Value src, mlir::Value dst)
 
   // 2. Try Reinterpret Copy if shape not match but numElements match
   if (src_type.getNumElements() == dst_type.getNumElements() && 
-      src_type.getElementType() == dst_type.getElementType()) {
+      src_type.getRank() != dst_type.getRank()) {
```
CeleNewYear: Delete

src/target/codegen_npuir_dev.cc
```diff
@@ -1087,62 +1087,125 @@ void CodeGenTileLangNPUIRDEV::SmartMemRefCopy(mlir::Value src, mlir::Value dst)
...
+        }
+    }
+
+    // 2. Compute strides for StridedLayoutAttr
+    // Rule: Calculate from the last dimension to the first. When encountering a dynamic dimension,
+    // the current and all preceding strides become dynamic.
+    llvm::SmallVector<int64_t> layout_strides(dst_type.getRank(), mlir::ShapedType::kDynamic);
+    int64_t current_stride = 1;
+    bool all_static_so_far = true;
+    bool current_stride_is_dynamic = false;
```
CeleNewYear: Maybe we don't need this，use break

PR #268, opened 2026-01-14
[feature]Erase manual workspace allocation C:V=1:1

zyh5913:


examples/sparse_flash_attention/example_sparse_flash_attn_erase_workspace.py
```diff
@@ -0,0 +1,321 @@
...
+            g_i = bz
+
+            s_i = (bx // REPLICATE_H)
+
+            H0 = g_i * padded_H + (0 if REPLICATE_H == 1 else (bx % REPLICATE_H) * 64)
+            H1 = H0 + H_per_block
+
+            T.copy(Q[b_i, s_i, H0:H1, :D], q_l1)
+            T.copy(Q[b_i, s_i, H0:H1, D:], q_tail_l1)
+            for _ in T.serial(NI):
```
xwhzz, line 155: With your method, it seems the AIC performs the first NI iteration, then the AIV performs the second. The order may not be strictly defined.

xwhzz: It’s not a well-formed program. Use a single iteration to handle all tile operations in an interleaved order.

PR #280, opened 2026-01-15
[AscendNPU-IR][Developer] Fix bug when one operand of binary op is scalar

dazuozcy:
As shown in `examples/vectorization_in_parallel.py`
``` python
for i in T.Parallel(block_N):
	C_VEC[i] = A_VEC[i] * B_VEC[2] + B_VEC[i]
```

To make `B_VEC[2]` codegen `f32` type (scalar type) instead of `memref<1xf32>` (vector type), in order to directly perform binary operations on vectors and scalars.

### DIFF
--- a/src/target/codegen_npuir_api.cc
+++ b/src/target/codegen_npuir_api.cc
@@ -1433,6 +1433,32 @@ void CodeGenTileLangNPUIRAPI::BitcastCodegen(const CallNode *op) {
   }
 }
 
+mlir::Value CodeGenTileLangNPUIRAPI::GenMemrefLoadFromRegion(const BufferLoadNode *op) {
+  auto buffer = op->buffer;
+  auto indices = op->indices;
+
+  // Check pre-conditions
+  if (op->dtype.lanes() != 1) {
+    LOG(FATAL) << "lanes not one";
+  }
+  if (op->dtype != buffer->dtype) {
+    LOG(FATAL) << "The load type and buffer element type do not match";
+  }
+
+  // Convert buffer from Buffer in TIR 2 memref in MLIR
+  auto mem = GetVarValue(buffer->data.get());
+
+  // Convert index from PrimExpr in TIR 2 index type in MLIR
+  SmallVector<mlir::Value> convert_inds;
+  for (auto index : indices) {
+    mlir::Value indexVal = CreateIndexCastOp(MakeValue(index));
+    convert_inds.push_back(indexVal);
+  }
+
+  // Create memef.load op in MLIR
+  return builder.create<mlir::memref::LoadOp>(builder.getUnknownLoc(), mem, convert_inds);
+}
+
 template <typename T>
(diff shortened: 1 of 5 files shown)


src/target/codegen_npuir_api.cc
```diff
@@ -1450,8 +1476,23 @@ void CodeGenTileLangNPUIRAPI::CreateHIVMBinaryVectorOp(const CallNode *op) {
...
+        const IntImmNode* int_imm = region_node->args[2 + i].as<IntImmNode>();
+        if (!int_imm || int_imm->value != 1) {
+          is_scalar_load = false;
+          break;
+        }
+      }
+      const IntImmNode* int_imm = region_node->args[2].as<IntImmNode>();
+      // If load only one element, do not use memref.subview, use memref.load as a scalar
+      if(is_scalar_load) {
+        src = GenMemrefLoadFromRegion(buffer_node);
```
baberrehman, line 1493: Why do we need this load here? Can we not directly add scalar with vector?

baberrehman (changes requested):

src/target/codegen_npuir_api.cc
baberrehman: Please add proper description of the issue/fix.

dazuozcy: > Why do we need this load here? Can we not directly add scalar with vector?

Sorry for noticing your comment so late. Let me explain your question below.
As you said, we can directly perform binary operations on vectors and scalars. 
But for the given scenario, as shown in examples/vectorization_in_parallel.py
``` python
for i in T.Parallel(block_N):
	C_VEC[i] = A_VEC[i] * B_VEC[2] + B_VEC[i]
```

Without the modification in this PR, the codegen of B_VEC[2] uses memref.subview and returns memref<1xf32> instead of a scalar.
Only when memref.load is used, a scalar type is returned.

dazuozcy: > Please add proper description of the issue/fix.

yes, description has been added.

baberrehman (approved):

src/target/codegen_npuir_dev.cc
```diff
@@ -1939,6 +1939,38 @@ void CodeGenTileLangNPUIRDEV::DotCodegen(const CallNode *op) {
...
+  // Convert index from PrimExpr in TIR 2 index type in MLIR
+  SmallVector<mlir::Value> convert_inds;
+  for (auto index : indices) {
+    mlir::Value indexVal = CreateIndexCastOp(MakeValue(index));
+    convert_inds.push_back(indexVal);
+  }
+
+  // Create memef.load op in MLIR
+  return builder.create<mlir::memref::LoadOp>(builder.getUnknownLoc(), mem, convert_inds);
+}
```
CeleNewYear, line 1972: We don't use memref.load in developer mode, because they are tensor

CeleNewYear (changes requested):

CeleNewYear (approved):
The memref.load operation in developer mode is treated as a known issue. And will be fixed latter

PR #293, opened 2026-01-17
[AscendNPU-IR][Expert] Support Dynamic shape

CeleNewYear:
1. Dynamic shape support in expert mode
2. The underscore (_) is allowed in kernel names.


dazuozcy:
LGTM

PR #304, opened 2026-01-19
[AscendNPU-IR][Expert][Developer] Support reshape op.

lzp1021:
Supports reshape operations in expert mode, including dynamic shapes.


testing/npuir/test_reshape.py
```diff
@@ -51,10 +73,12 @@ def generate_tensor(shape, dtype, clear=False):
...
 
 def main(main_args):
-    os.environ["TILELANG_ASCEND_MODE"] = "dev"
     M = main_args.M
     N = main_args.N
-    func = reshape_demo(M, N)
+    if os.environ['TILELANG_ASCEND_MODE'] == 'Expert':
+        func = reshape_demo_exp(M, N)
+    else:
+        func = reshape_demo_dev(M, N)
```
baberrehman: This will call `reshape_demo_dev(M, N)` if `TILELANG_ASCEND_MODE` is not set (which is different from TileLang behavior). Please change the condition as:

```
    if os.environ['TILELANG_ASCEND_MODE'] == 'Dev':
        func = reshape_demo_dev(M, N)
    else:
        func = reshape_demo_exp(M, N)
```

src/target/codegen_npuir_api.cc
```diff
@@ -1589,6 +1589,59 @@ void CodeGenTileLangNPUIRAPI::DebugPrintCodegen(const CallNode *op) {
                                        mlir::hivm::TCoreTypeAttr{});
 }
 
+void CodeGenTileLangNPUIRAPI::ReshapeCodegen(const CallNode *op) {
+  tvm::tl::NpuirReshape npuirop(op->args, this->vmap);
+  Value src = GenSubviewFromRegion(npuirop.src, npuirop.src_range);
+  Value dst = GenSubviewFromRegion(npuirop.dst, npuirop.dst_range);
+
+  auto srcTy = src.getType().cast<mlir::MemRefType>();
+  auto elemTy = srcTy.getElementType();
```
baberrehman, line 1640: I see only one use of `elemTy`. Please remove this variable and directly use `srcTy.getElementType()` at line 1617.

src/target/codegen_npuir_api.cc
```diff
@@ -1589,6 +1589,59 @@ void CodeGenTileLangNPUIRAPI::DebugPrintCodegen(const CallNode *op) {
...
+
+  ArrayRef<int64_t> dstShape = npuirop.dst_shape;
+  
+  SmallVector<int64_t> stridesVec;
+  int64_t stride = 1;
+  for (int i = dstShape.size() - 1; i >= 0; --i) {
+    stridesVec.push_back(stride);
+    stride *= dstShape[i];
+  }
+  std::reverse(stridesVec.begin(), stridesVec.end());
```
baberrehman, line 1652: Why do we need to call `reverse` on `stridesVec`?

baberrehman (dismissed):

src/target/codegen_npuir_api.cc
lzp1021: Because the stride is calculated from the last dimension forward.

testing/npuir/test_reshape.py
lzp1021: Thanks, I’ve already fixed it.

src/target/codegen_npuir_api.cc
lzp1021: Thanks, I’ve already fixed it.

testing/npuir/test_reshape.py
```diff
@@ -14,7 +14,29 @@
...
+            a = T.alloc_ub((M, N), dtype)
+            tmp = T.alloc_ub((N, M), dtype)
+            b = T.alloc_ub((N, M), dtype)
+            c = T.alloc_ub((N, M), dtype)
+            T.copy(A, a)
+            T.copy(B, b)
+
+            T.npuir_reshape(a, tmp)
+            T.npuir_add(b, tmp, c)
+            T.copy(c, C)
```
CeleNewYear: Add a scenario where the tail block needs to be processed to verify whether reshape takes effect in dynamic shape scenarios.

CeleNewYear (changes requested):
Please consider whether the reshape implementation is reliable in dynamic shape scenarios.

src/target/codegen_npuir_api.cc
baberrehman: Thank you.

testing/npuir/test_reshape.py
baberrehman: Thanks.

CeleNewYear (approved):
LGTM

PR #306, opened 2026-01-19
[AscendNPU-IR][Developer]Support Slice in ReduceOp

dangyw18:
Support Slice in ReduceOp in Dev mode
eg: T.reduce(A, row_max, dims=1, size=[4,4], clear=True) where A is (4,8) and only A[:,:4] is calculated in reduce.


baberrehman (changes requested):
Please update description and clearly specify the two features this MR supports:
1) Sigmoid
2) Support slice in vreduce

PR #310, opened 2026-01-20
[AscendNPU-IR][Expert] support reduce op

james1314520:
support reduce op for loop


testing/npuir/test_pipelined_reduce.py
```diff
@@ -0,0 +1,64 @@
...
+
+    print("ref_output")
+    print(ref_output)
+
+    torch.testing.assert_close(output, ref_output, rtol=1e-2, atol=1e-2)
+    print("\033[92mAll check passed!\033[0m")
+
+
+if __name__ == "__main__":
+    os.environ['TILELANG_ASCEND_MODE'] = 'Developer'
```
baberrehman, line 62: Please print the log here:
`print("Running in developer mode")`

baberrehman (changes requested):

PR #312, opened 2026-01-20
[AscendNPU-IR][Developer]Support Slice in ReduceOp

dangyw18:
Support Slice in VReduceOp in Dev mode

This MR enchances the 'tl.npuir_reduce' codegen to support two key features:
1. **Partial tensor reduction**: Reduce operations can now be applied to a sub-region(slice) of the source tensor, specified by 'src_range' (min/extent per dimension).
2. Hybrid static/dynamic range building: The offset and size of the reduction region can be either compile-time constants or runtime values, enabling flexible kernel generation for both static and dynamic shapes.
 
The implementation inserts a 'tensor.extract_slice' before the 'hivm.vreduce' op to materialize the input view, ensuring correct memory access on Asceng NPU.

Usage example:  T.reduce(A, row_max, dims=1, size=[4,4], clear=True) where A is (4,8) but only A[:,:4] is calculated in reduce_max.


baberrehman (dismissed):
Please add a test case.

PR #313, opened 2026-01-20
[AscendNPU-IR][Expert] support exp ln sqrt rsqrt rec not op for loop

james1314520:
support exp ln sqrt rsqrt rec not op for loop


baberrehman (changes requested):
How many ops we need to support we need to support for `T.pipelined`? I see a few MRs for this purpose with a few ops. We should support all in one instead of doing incremental work.

testing/npuir/test_pipelined.py
```diff
@@ -172,6 +172,230 @@ def test_for_div():
...
+
+    print("ref_output")
+    print(input)
+
+    torch.testing.assert_close(output, input, rtol=1e-2, atol=1e-2)
+    print("\033[92mAll check passed!\033[0m")
+
+
 if __name__ == "__main__":
     os.environ['TILELANG_ASCEND_MODE'] = 'Developer'
```
baberrehman: Please add a log here: `print("Running in developer mode")`

baberrehman (changes requested):

PR #315, opened 2026-01-20
[Example] a convolution example

BikingNow:
a convolution example

### DIFF
--- a/examples/convolution/example_convolution.py
+++ b/examples/convolution/example_convolution.py
@@ -1,0 +1,127 @@
+import torch
+import torch.nn.functional as F
+
+import tilelang
+import tilelang.language as T
+import torch
+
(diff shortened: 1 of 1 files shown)


examples/convolution/example_convolution.py
```diff
@@ -0,0 +1,127 @@
...
+                    T.copy(B[k * block_K, by * block_N], B_L1)
+
+                    T.gemm_v0(A_L1, B_L1, C_L0, init=(k == 0))
+
+                T.copy(C_L0, C[bx * block_M, by * block_N])
+
+    return main
+
+
+def im2col(input_tensor: torch.Tensor, KH: int, KW: int,
```
benyang0506, line 54: Need to encapsulate im2col as a primitive in the future, or use T.Parallel to support similar functionality.

PR #348, opened 2026-01-23
[AscendNPU-IR][Expert] npuir_brc supporting variable parameters.

lzp1021:
[AscendNPU-IR][Expert] npuir_brc supporting variable parameters.

### DIFF
--- a/src/target/codegen_npuir_api.cc
+++ b/src/target/codegen_npuir_api.cc
@@ -196,31 +196,6 @@ getBroadcastDim(const llvm::ArrayRef<long int> &buffer_shape0,
   return dims;
 }
 
-static llvm::SmallVector<int64_t>
-getBroadcastDim(const llvm::ArrayRef<int64_t> buffer_shape0,
-                const Array<PrimExpr> &buffer_shape1) {
-  llvm::SmallVector<int64_t> dims;
-  if (buffer_shape0.empty() || buffer_shape1.empty()) {
-    return dims;
-  }
-  CHECK(buffer_shape0.size() == buffer_shape1.size());
-  for (int i = 0; i < buffer_shape0.size(); i++) {
-    auto *b1 = as_const_int(buffer_shape1[i]);
-    CHECK(b1) << "buffer_shape1[" << i << "] is not a const int";
-
-    if (buffer_shape0[i] == 1 &&
-        *b1 != 1) {
-      dims.emplace_back(i);
-    } else if (buffer_shape0[i] != 1 &&
-               *b1 == 1) {
-      dims.emplace_back(i);
-    } else {
-      CHECK(buffer_shape0[i] == *b1);
-    }
-  }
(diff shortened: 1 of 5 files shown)


src/target/codegen_npuir_api.cc
```diff
@@ -1067,7 +1067,7 @@ void CodeGenTileLangNPUIRAPI::VbrcCodegen(const CallNode *op) {
   tvm::tl::NpuirBrc npuirop(op->args, this->vmap);
   mlir::Value src;
   llvm::ArrayRef<int64_t> inBufferShape;
-  if (npuirop.in.as<IntImm>() || npuirop.in.as<FloatImm>()) {
+  if (npuirop.in.as<tvm::tir::Var>()) {
     // Scalar case
```
baberrehman, line 1075: Does it mean that we aren't handling constants now?

baberrehman (dismissed):

src/target/codegen_npuir_api.cc
lzp1021: No, this modification still supports constants.

src/target/codegen_npuir_api.cc
```diff
@@ -1067,7 +1067,10 @@ void CodeGenTileLangNPUIRAPI::VbrcCodegen(const CallNode *op) {
   tvm::tl::NpuirBrc npuirop(op->args, this->vmap);
   mlir::Value src;
   llvm::ArrayRef<int64_t> inBufferShape;
-  if (npuirop.in.as<IntImm>() || npuirop.in.as<FloatImm>()) {
+  bool isScalar =
+      !npuirop.in.as<tvm::tir::Buffer>() &&
+      !npuirop.in.as<tvm::tir::BufferRegion>();
+  if (isScalar) {
```
CeleNewYear, line 1074: BufferLoad Node need to be test. I think there is a problem with the handling here.

src/target/codegen_npuir_api.cc
```diff
@@ -1084,7 +1113,7 @@ void CodeGenTileLangNPUIRAPI::VbrcCodegen(const CallNode *op) {
   if (!inBufferShape.empty()) {
     auto outMemref = llvm::dyn_cast<TypedValue<MemRefType>>(dst);
     auto outBufferShape = outMemref.getType().getShape();
-    auto broadcastDim = getBroadcastDim(npuirop.src->shape, npuirop.dst->shape);
+    auto broadcastDim = getBroadcastDim(inBufferShape, npuirop.dst->shape);
```
CeleNewYear: Why only change npuirop.src->shape，but not npuirop.dst->shape

src/target/codegen_npuir_api.cc
lzp1021: I've tested and this is ok.

CeleNewYear (approved):
LGTM

PR #353, opened 2026-01-24
[AscendNPU-IR][Developer] Refact Copy Codegen

mmyy1217:
**Generates MLIR for `tl.ascend_copy`, handling data movement between Global Memory (GM) and local Unified Buffer (UB), with support for rank reduction and dynamic shapes.**

### Overview

This lowering dispatches based on the source/destination IR types (MemRef vs Tensor) and emits an explicit MLIR sequence per case. The core design is to compute a *canonical copy shape* by dropping static-1 dimensions, enabling rank-reduced `memref.subview` while preserving destination tensor semantics via reshape/cast.

#### 1) Load (MemRef -> Tensor)

* **Concept:** Copy from GM (MemRef) into local UB, then materialize as a Tensor.
* **Rank/Shape Handling:**

  1. Canonicalize copy rank by dropping static-1 dims from the **source range**.
  2. Create a rank-reduced `memref.subview` of the source GM to match the canonical rank.
  3. Allocate a static local UB memref using the destination tensor shape with static-1s removed.
  4. Create an UB `memref.subview` matching the canonical rank.
  5. `memref.copy` from GM view to UB view.
  6. Convert UB to tensor via `bufferization.to_tensor`.
  7. Reshape/cast to match the destination slice type, then `tensor.insert_slice` into the destination tensor (and update the var binding).

#### 2) Store (Tensor -> MemRef)

* **Concept:** Store a tensor slice back to GM.
* **Logic:**

  1. Extract a slice from the source tensor (`tensor.extract_slice`).
  2. Create a destination GM subview (`memref.subview`).
  3. Resolve rank/shape mismatches via reshape, and type mismatches via cast.
  4. Write using `bufferization.materialize_in_destination` (writable).

#### 3) Copy (Tensor -> Tensor)

* **Concept:** Tensor-to-tensor movement / layout manipulation through slice extraction and insertion.
* **Logic:**

  1. Extract slices from source and destination tensors.
  2. Reshape/cast the source slice to match the destination slice type.
  3. Insert into destination using `tensor.insert_slice` (and update the var binding).

### Example Flow (Load)

```mlir
// Input:  tl.ascend_copy(src_gm[...], dst_tensor[...])
// Output:
%src_view = memref.subview %src_gm [...] : (rank-reduced)
%ub       = memref.alloc()              : (local UB)
memref.copy %src_view, %ub
%val      = bufferization.to_tensor %ub
%result   = tensor.insert_slice %val into %dst_tensor [...]
```

# Contributions：
- Zuoming: alloc, memref copy, to_tensor
- Wenqi: reshape
- Chenyu: dtypeCast, insert_slice, bufferization::to_tensor

### DIFF
--- a/examples/flash_attn_npuir_dev.py
+++ b/examples/flash_attn_npuir_dev.py
@@ -86,7 +86,7 @@ def online_flash_attention(block_M, block_N, block_K, dtype="float16", accum_dty
             O_cast = T.alloc_shared([block_m, dim], dtype)
             T.vcast(acc_o, O_cast, round_mode="rint")
             real_m = T.min(block_m, seq_len - cid * block_m)
-            T.copy(O_cast, Output[cid * block_m : (cid+1) * block_m, :],size=[real_m, dim])
+            T.copy(O_cast, Output[cid * block_m : cid * block_m + real_m, :],size=[real_m, dim])
 
     return flash_attention
 
--- a/examples/flash_attn_npuir_dev.py
+++ b/examples/flash_attn_npuir_dev.py
@@ -86,7 +86,7 @@ def online_flash_attention(block_M, block_N, block_K, dtype="float16", accum_dty
             O_cast = T.alloc_shared([block_m, dim], dtype)
             T.vcast(acc_o, O_cast, round_mode="rint")
             real_m = T.min(block_m, seq_len - cid * block_m)
-            T.copy(O_cast, Output[cid * block_m : cid * block_m + real_m, :],size=[real_m, dim])
+            T.copy(O_cast, Output[cid * block_m : (cid+1) * block_m, :],size=[real_m, dim])
 
     return flash_attention
 
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -889,13 +889,13 @@ inline void CodeGenTileLangNPUIRDEV::UpdatePrimExprMap(const PrimExprNode * key,
 mlir::Value CodeGenTileLangNPUIRDEV::CreateCastIfTypeMismatch(mlir::Value src, mlir::Value dst) {
   // src is always a tensor, dst may be a tensor or a memref, the return value is always a tensor
 
-  auto srcTensorTy = src.getType().cast<mlir::TensorType>(); 
-  assert(srcTensorTy && "src must be a tensor");
+  auto srcTensorTy = src.getType().dyn_cast<mlir::TensorType>();
+  ICHECK(srcTensorTy) << "src must be a tensor";
 
-  mlir::Type srcElemTy  = mlir::getElementTypeOrSelf(src.getType());
-  mlir::Type dstElemTy  = mlir::getElementTypeOrSelf(dst.getType());
+  mlir::Type srcElemTy = mlir::getElementTypeOrSelf(src.getType());
+  mlir::Type dstElemTy = mlir::getElementTypeOrSelf(dst.getType());
 
-  if (srcElemTy  == dstElemTy ) {
+  if (srcElemTy == dstElemTy ) {
     return src;
   }
 
@@ -1032,6 +1032,78 @@ mlir::Value CodeGenTileLangNPUIRDEV::MaybeReshapeTensor(mlir::Value src, mlir::V
       shape_tensor);
   
   return reshaped_tensor;
+}
+
+mlir::Value CodeGenTileLangNPUIRDEV::MaybeReshapeTensorByDstSize(
+    mlir::Value src,
+    llvm::ArrayRef<mlir::OpFoldResult> sizes) {
+
+  ICHECK(src.getType().isa<mlir::TensorType>()) << "src must be a tensor";
+  auto srcTensorTy = src.getType().cast<mlir::TensorType>();
+  auto loc = builder.getUnknownLoc();
+
+  llvm::SmallVector<mlir::Value> targetShapeVals;     // runtime values
+  llvm::SmallVector<int64_t>     targetShapeStatic;   // static shape (kDynamic)
+
+  for (mlir::OpFoldResult ofr : sizes) {
+    if (auto attr = ofr.dyn_cast<mlir::Attribute>()) {
+      auto intAttr = attr.cast<mlir::IntegerAttr>();
+      int64_t dim = intAttr.getInt();
+
+      targetShapeStatic.push_back(dim);
+      targetShapeVals.push_back(
+          builder.create<mlir::arith::ConstantIndexOp>(loc, dim));
+
+    } else if (auto val = ofr.dyn_cast<mlir::Value>()) {
+      targetShapeStatic.push_back(mlir::ShapedType::kDynamic);
+      targetShapeVals.push_back(val);
+
+    } else {
+      llvm_unreachable("Invalid OpFoldResult in sizes");
+    }
+  }
+
+  if (auto rankedTy = srcTensorTy.dyn_cast<mlir::RankedTensorType>()) {
+    if (rankedTy.getRank() == (int64_t)targetShapeStatic.size()) {
+      bool same = true;
+      for (int i = 0; i < rankedTy.getRank(); ++i) {
+        int64_t s0 = rankedTy.getDimSize(i);
+        int64_t s1 = targetShapeStatic[i];
+        if (s0 != s1) {
+          if (mlir::ShapedType::isDynamic(s0) ||
+              mlir::ShapedType::isDynamic(s1)) {
+            same = false;
+            break;
+          }
+          same = false;
+          break;
+        }
+      }
+      if (same) return src;
+    }
+  }
+
+  auto targetTensorTy = mlir::RankedTensorType::get(
+      targetShapeStatic,
+      srcTensorTy.getElementType());
+
+  mlir::Value shapeTensor;
+  {
+    auto shapeTensorTy = mlir::RankedTensorType::get(
+        {static_cast<int64_t>(targetShapeVals.size())},
+        builder.getIndexType());
+
+    shapeTensor = builder.create<mlir::tensor::FromElementsOp>(
+        loc, shapeTensorTy, targetShapeVals);
+  }
+
+  auto reshaped = builder.create<mlir::tensor::ReshapeOp>(
+      loc,
+      targetTensorTy,
+      src,
+      shapeTensor);
+
+  return reshaped;
 }
 
 // Converts a TileLang range/region descriptor into {offs, sizes, strides}.
@@ -1177,9 +1249,9 @@ void CodeGenTileLangNPUIRDEV::EmitCopyMemrefToTensor(
   auto src_memref_type_ori = src.getType().cast<mlir::MemRefType>();
 
   // Always form dst_slice first
-  mlir::Value dst_slice = builder.create<mlir::tensor::ExtractSliceOp>(
-      loc, dst, dstR.offs, dstR.sizes, dstR.strides);
-  auto dst_slice_ty = dst_slice.getType().cast<mlir::RankedTensorType>();
+  // mlir::Value dst_slice = builder.create<mlir::tensor::ExtractSliceOp>(
+  //     loc, dst, dstR.offs, dstR.sizes, dstR.strides);
+  // auto dst_slice_ty = dst_slice.getType().cast<mlir::RankedTensorType>();
 
   // 1) Canonicalize copy rank: drop static-1 dims using src_sizes
   CollapsedDims srcC = CollapseStaticOneDims(srcR.sizes);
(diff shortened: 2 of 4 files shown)


examples/flash_attn_npuir_dev.py
```diff
@@ -86,7 +86,7 @@ def flash_attention(
             O_cast = T.alloc_shared([block_m, dim], dtype)
             T.vcast(acc_o, O_cast, round_mode="rint")
             real_m = T.min(block_m, seq_len - cid * block_m)
-            T.copy(O_cast, Output[cid * block_m : (cid+1) * block_m, :],size=[real_m, dim])
+            T.copy(O_cast, Output[cid * block_m : cid * block_m + real_m, :],size=[real_m, dim])
```
CeleNewYear: We need a new PR to completely modify the fa operator in developer mode. In the current open source version, the tail block is considered when store out, but not considered when copy in.

I think here, we can remove real_m

tilelang/jit/jit_npu.py
```diff
@@ -1267,21 +1267,21 @@ def _npuir_to_bin_enable_npu_compile(self):
...
-            print("AscendNPU IR OPT success:", ret.stdout)
+            # opt_cmd_list = (
+            #     [npu_compiler_opt_path, ttadapter_path]
+            #     + _opt_option_list
+            #     + ["-o", ttadapter_path]
+            # )
+            # ret = subprocess.run(
+            #     opt_cmd_list, capture_output=True, check=True, text=True
+            # )
+            # print("AscendNPU IR OPT success:", ret.stdout)
```
CeleNewYear: We need a complete solution to resolve the conflict between CANN 8.3 and 8.5, which should be a complete PR. In this PR, no modification is required here.

src/target/codegen_npuir_dev.cc
```diff
@@ -415,12 +415,19 @@ void CodeGenTileLangNPUIRDEV::VisitStmt_(const tir::ForNode *op) {
...
+  // Any tensor degined outside should be passed as iter_args
+  for (const auto& layer : this->var_map_) {
+    for (const auto& entry : layer) {
+      const tir::VarNode* var = entry.first;
+      mlir::Value val = entry.second;
+      if (val && val.getType().isa<mlir::TensorType>()) {
+        loop_carried_vars.insert(var);
+      }
+    }
+  }
```
CeleNewYear: When our operators are complex and contain multiple for loops, it will make the generated MLIR less user-friendly. Change it back. Just input what we use.

src/target/codegen_npuir_dev.cc
```diff
@@ -884,139 +891,48 @@ inline void CodeGenTileLangNPUIRDEV::UpdatePrimExprMap(const PrimExprNode * key,
   this->prim_expr_map[{GetRef<PrimExpr>(key), curr_block}] = val;
 }
 
-// Convert tensor value to its underlying memref
-mlir::Value CodeGenTileLangNPUIRDEV::ConvertTensorToMemref(mlir::Value value) {
-  // Case 1: tensor.empty -> memref.alloc
-  if (auto emptyOp = value.getDefiningOp<mlir::tensor::EmptyOp>()) {
+/// Helper functions for AscendCopyCodegen
+// -------------------------------------------------------------
```
CeleNewYear: Delete. Not only for copy, but also for other ops.

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
 
-    llvm::SmallVector<mlir::OpFoldResult> offsets;
-    offsets.push_back(offsetValue);
+// Reshape tensor
+mlir::Value CodeGenTileLangNPUIRDEV::MaybeReshapeTensor(mlir::Value src, mlir::Value dst) {
 
-    // 1. get sizes for dst value.
-    llvm::SmallVector<mlir::OpFoldResult> sizes;
-    llvm::SmallVector<mlir::Value> dim_values;
+  auto src_tensor_type = src.getType().cast<mlir::TensorType>();
```
CeleNewYear, line 957: Add a note to indicate that the input can only be a tensor; at the same time, add a protection here to report an error when the input is not a tensor.

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
-      if (mlir::ShapedType::isDynamic(dim_size)) {
-        all_static_so_far = false;
-        break;
+  } else if (dst.getType().isa<mlir::RankedTensorType>()) {
+    auto dst_type = dst.getType().cast<mlir::RankedTensorType>();
+    // Iterate through destination memref dimensions
+    for (int i = 0; i < dst_type.getRank(); ++i) {
+      int64_t dim = dst_type.getDimSize(i);
+      if (mlir::ShapedType::isDynamic(dim)) {
+        // Dynamic dimension: get actual size at runtime using memref.dim
```
CeleNewYear: using tensor.dim

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
+  r.sizes = std::move(sizes);
+  r.strides = std::move(strides);
+  return r;
+}
+
+// Allocates a statically-shaped local UB memref with a fixed alignment.
+mlir::Value CodeGenTileLangNPUIRDEV::CreateStaticLocalUB(
+    llvm::ArrayRef<int64_t> shape, mlir::Type elem_type, mlir::Location loc) {
+  auto memrefType = mlir::MemRefType::get(shape, elem_type);
+  mlir::IntegerAttr alignmentAttr = builder.getI64IntegerAttr(64);
```
CeleNewYear: Please explain the meaning of the magic number 64

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
+
+  for (int64_t d : dst_tensor_type_ori.getShape()) {
+    if (d == 1) continue;  // drop static 1 as in original
+    ub_alloc_shape.push_back(d);
+  }
+  if (ub_alloc_shape.empty()) ub_alloc_shape.push_back(1);
+  return ub_alloc_shape;
+}
+
+// Emitters
```
CeleNewYear: what this means

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
+    mlir::Value src, mlir::Value dst,
+    const SliceRange& srcR, const SliceRange& dstR,
+    mlir::Location loc) {
+  auto dst_tensor_type_ori = dst.getType().cast<mlir::RankedTensorType>();
+  auto src_memref_type_ori = src.getType().cast<mlir::MemRefType>();
+
+  // Always form dst_slice first
+  mlir::Value dst_slice = builder.create<mlir::tensor::ExtractSliceOp>(
+      loc, dst, dstR.offs, dstR.sizes, dstR.strides);
+  auto dst_slice_ty = dst_slice.getType().cast<mlir::RankedTensorType>();
```
CeleNewYear: Why do we need this

CeleNewYear: when copy memref to tensor, if the shape don't match, we can just use tensor.empty and insert_slice.  There is no need to extract the origional value

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
+  llvm::ArrayRef<mlir::OpFoldResult> copy_sizes = srcC.sizes;
+  llvm::ArrayRef<int64_t> copy_projected = srcC.projected;
+
+  // 2) Build src_view as rank-reduced subview to copy rank
+  ICHECK((int64_t)srcR.offs.size() == src_memref_type_ori.getRank());
+  ICHECK((int64_t)srcR.sizes.size() == src_memref_type_ori.getRank());
+  ICHECK((int64_t)srcR.strides.size() == src_memref_type_ori.getRank());
+
+  auto reduced_src_type =
+      mlir::memref::SubViewOp::inferRankReducedResultType(
```
CeleNewYear: there are two inferRankReducedResultType, can reuse?

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
+      loc, ub_view, /*restrict=*/true, /*writable=*/false);
+
+  // 7) Reshape to dst_slice if needed
+  mlir::Value reshaped_tensor;
+  auto loaded_ty = loaded_tensor.getType().cast<mlir::RankedTensorType>();
+  if (loaded_ty.getShape() != dst_slice_ty.getShape()) {
+    reshaped_tensor = MaybeReshapeTensor(loaded_tensor, dst_slice);
+  } else {
+    reshaped_tensor = loaded_tensor;
   }
```
CeleNewYear: This part should be completely completed by MaybeReshapeTensor. If the shapes are the same, the original tensor is returned directly. Otherwise, the reshaped tensor is returned. 

Which means delete if else

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
-  ICHECK(false) << "SmartMemRefCopy: Shape mismatch and cannot interpret cast. ";
+
+  // 8) Type Cast
+  mlir::Value casted_tensor = CreateCastIfTypeMismatch(reshaped_tensor, dst_slice);
```
CeleNewYear: For the reshape above, you can refer to here

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
+
+  auto src_tensor_type = src_slice.getType().cast<mlir::RankedTensorType>();
+  auto dst_memref_type = dst_view.getType().cast<mlir::MemRefType>();
+
+  mlir::Value reshaped_tensor;
+  if (src_tensor_type.getShape() != dst_memref_type.getShape()) {
+    reshaped_tensor = MaybeReshapeTensor(src_slice, dst_view);
+  } else {
+    reshaped_tensor = src_slice;
   }
```
CeleNewYear: same as before

src/target/codegen_npuir_dev.cc
```diff
@@ -1047,200 +963,469 @@ CodeGenTileLangNPUIRDEV::CreateOpFoldResultArray(const Array<Range>& range) {
...
+  auto dst_tensor_type = dst_slice.getType().cast<mlir::RankedTensorType>();
+
+  mlir::Value reshaped_tensor;
+  if (src_tensor_type.getShape() != dst_tensor_type.getShape()) {
+    reshaped_tensor = MaybeReshapeTensor(src_slice, dst_slice);
   } else {
-    // Insert at block start for arguments.
-    builder.setInsertionPointToStart(memref.getParentBlock());
+    reshaped_tensor = src_slice;
   }
```
CeleNewYear: same as before

CeleNewYear (changes requested):
It's generally good, but some details need to be revised.

src/target/codegen_npuir_dev.cc
mmyy1217: done

tilelang/jit/jit_npu.py
baberrehman: This has been merged in a separate PR. Please remove these changes from this PR.

src/target/codegen_npuir_dev.cc
mmyy1217: reused

src/target/codegen_npuir_dev.cc
```diff
@@ -884,361 +885,506 @@ inline void CodeGenTileLangNPUIRDEV::UpdatePrimExprMap(const PrimExprNode * key,
...
 
-    auto memrefType = mlir::MemRefType::get(
-      tensorType.getShape(),
-      tensorType.getElementType()
-    );
-    
-    mlir::Location loc = emptyOp.getLoc();
-    mlir::OpBuilder::InsertionGuard guard(builder);
+  auto srcTensorTy = src.getType().cast<mlir::TensorType>(); 
+  assert(srcTensorTy && "src must be a tensor");
```
CeleNewYear: This code will not work as you expect. cast is a forced cast; you should use dyn_cast here instead.

CeleNewYear (changes requested):

src/target/codegen_npuir_dev.cc
```diff
@@ -884,361 +885,506 @@ inline void CodeGenTileLangNPUIRDEV::UpdatePrimExprMap(const PrimExprNode * key,
...
-  
-  // Case 2: bufferization.to_tensor -> extract original memref
-  if (auto toTensorOp = value.getDefiningOp<mlir::bufferization::ToTensorOp>()) {
-    return toTensorOp.getMemref();
-  }
-  
-  // Case 3: already a memref
-  if (value.getType().isa<mlir::MemRefType>()) {
-    return value;
+  if (srcElemTy  == dstElemTy ) {
```
CeleNewYear: There is an extra space after dstElemTy here.

src/target/codegen_npuir_dev.cc
mmyy1217: Removed the extract slice both here and in tensor->tensor dst. Use a new maybereshape that collect dst_shape instead of dst here.

CeleNewYear (approved):
lGTM

PR #355, opened 2026-01-26
[AscendNPU-IR][Developer] Fix bugs related to T.Parallel

voidLitchi:
This MR contains two submits:

1. Distinguish between exp and vexp to support scenarios where exp appears in T.Parallel
2. Fix bug in T.Parallel that occurs when performing binary operations between vector and scalar

The relevant scenario has been validated and passed

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -1865,38 +1865,6 @@ void CodeGenTileLangNPUIRDEV::DotCodegen(const CallNode *op) {
   SetVarValue(npuirop.dst, newMmadL1OpValue);
 }
 
-mlir::Value CodeGenTileLangNPUIRDEV::GenMemrefLoadFromRegion(const BufferLoadNode *op) {
-  auto buffer = op->buffer;
-  auto indices = op->indices;
-
-  // Check pre-conditions
-  if (op->dtype.lanes() != 1) {
-    LOG(FATAL) << "lanes not one";
-  }
-  if (op->dtype != buffer->dtype) {
-    LOG(FATAL) << "The load type and buffer element type do not match";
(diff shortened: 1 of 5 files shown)


CeleNewYear (approved):
LGTM

tilelang/language/__init__.py
```diff
@@ -88,7 +88,7 @@
     npuir_shl,
     npuir_shr,
     npuir_exp,
-    npuir_exp as exp,
+    # npuir_exp as exp,
```
dazuozcy: just delete unused code

src/target/codegen_npuir_dev.cc
```diff
@@ -1942,9 +1942,10 @@ void CodeGenTileLangNPUIRDEV::CreateHIVMBinaryVectorOp(const CallNode *op) {
       const IntImmNode* int_imm = region_node->args[2].as<IntImmNode>();
       // If load only one element, do not use memref.subview, use memref.load as a scalar
       if(is_scalar_load) {
-        src = GenMemrefLoadFromRegion(buffer_node);
```
dazuozcy, line 1945: delete the definition of this function if it's no longer used in this file

dazuozcy (changes requested):

dazuozcy (approved):
LGTM

src/target/codegen_npuir_dev.cc
voidLitchi: As your comment, function 'GenMemrefLoadFromRegion' has been deleted.

tilelang/language/__init__.py
voidLitchi: All right, the annotation has been deleted.

PR #359, opened 2026-01-27
[Refactor]extract operation config into common configuration

jackwangc:


src/transform/common/operation_config.h
```diff
@@ -0,0 +1,232 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
benyang0506, line 2: This file currently has an Apache 2.0 license header, but TileLang projects should use the MIT license. Please update the header to match the standard MIT license header found in other Tilelang Ascend pass (Copyright Tile-AI Corporation).

PR #368, opened 2026-01-28
[AscendNPU-IR][developer] Fix vector broadcast

happyhappyhappy-arch:
Fix Ascend NPU IR codegen for vector broadcast (e.g. `A[:,:] + B[0,:]`) by using region extent instead of buffer shape when building the slice, and add `vec_add_auto_brc.py` as a regression example.


CeleNewYear (approved):
LGTM

PR #369, opened 2026-01-28
[AscendNPU-IR][Developer] Support to specify the region in operands of matmul op

voidLitchi:
Modifed 'CodeGenTileLangNPUIRDEV::DotCodegen' to support support specifying regions for operands in matrix multiplication.

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -2076,8 +2076,8 @@ void CodeGenTileLangNPUIRDEV::DotCodegen(const CallNode *op) {
 
   mlir::Location unknown_loc = builder.getUnknownLoc();
   mlir::IndexType idx_ty = builder.getIndexType();
-  mlir::Value a = GetVarValue(npuirop.src0);
-  mlir::Value b = GetVarValue(npuirop.src1);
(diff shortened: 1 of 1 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -2076,8 +2076,8 @@ void CodeGenTileLangNPUIRDEV::DotCodegen(const CallNode *op) {
 
   mlir::Location unknown_loc = builder.getUnknownLoc();
   mlir::IndexType idx_ty = builder.getIndexType();
-  mlir::Value a = GetVarValue(npuirop.src0);
-  mlir::Value b = GetVarValue(npuirop.src1);
+  mlir::Value a = GenExtractSliceFromRegion(npuirop.src0);
+  mlir::Value b = GenExtractSliceFromRegion(npuirop.src1);
```
baberrehman: Is this a generic extract slice on tensors? Same as subview for memref?

CeleNewYear (approved):
LGTM

src/target/codegen_npuir_dev.cc
CeleNewYear: Yes, they are the same

PR #384, opened 2026-01-29
[AscendNPU-IR][developer] Fix reduce issue

taowenqi:
After merging with Baber's NPUIR update, it encountered reduce-related issues in developer mode. Specifically, the new version of hivm.hir.vreduce does not support inconsistencies in non-reduce dimensions. The solution is to insert extract_slice and insert_slice before and after vreduce respectively to ensure consistency in non-reduce axis dimensions during vreduce.


CeleNewYear (approved):
LGTM

PR #406, opened 2026-01-31
[AscendNPU-IR][Developer] Support slice Operation for reduce and copy Op

shakespearjia:
Reduce Op and copy Op are now support both slice form and size specification form, and I think it's reasonable to keep both.
take this copy for example:
- T.copy(Input[offset_m, offset_n], src, size=[remain_m, remain_n])

if we convert it into slice form, it would be:
- T.copy(Input[offset_m: offset_m + remain_m, offset_n: offset_n + remain_n], src)

it would be too long


CeleNewYear (approved):
LGTM

PR #439, opened 2026-02-06
[bugfix]Modify the kv_group adaptation in the sparse_flash_attn_gqa_pipeline.py test case to PTO

kangjiaming1:
Modify the kv_group adaptation in the sparse_flash_attn_gqa_pipeline.py test case to PTO.Due to the limitation of a minimum classification of 16 for L0 and L1 at the bottom level of PTO, the classification of L0 and L1 in the current use case is 8. When running PTO, an error occurs during compilation. Therefore, it is necessary to first modify the kv_group of the use case to adapt to the minimum classification limitation of PTO. We will adapt the scene through PASS in the future.


kangjiaming1:
Closes #440

PR #443, opened 2026-02-06
[AscendNPU-IR][Expert] support sigmoid in codegen

dangyw18:
support sigmoidOp in exp with the same method as dev mode.


dazuozcy:
LGTM

PR #456, opened 2026-02-11
[AscendNPU-IR][Expert] Support rank-mismatch and dynamic shape in nd2nz and fixpipe

mmyy1217:
1. Rank reduction for nd2nz / fixpipe GM operands.
Introduce GenRankReducedSubviewFromRegion to drop static-1 dimensions from Region slices, so that higher-dimensional GM shapes (e.g. 1×M×N from a 3D/4D tensor) can be properly consumed by 2D Cube operations.
2. min_rank constraint for consistent callee signatures.
Add a min_rank parameter (default 2 for GM operands) to prevent aggressive rank reduction from collapsing shapes like [1, N] to 1D while other slices remain 2D. 
This ensures all GM operands within the same function share a uniform rank, avoiding type mismatches when convert-hivm-to-std outlines them into a unified callee signature.
3. Test coverage for sliced and dynamic-shape Cube pipelines.
Add test_copy_sliced_cube.py (static sliced copy with 2D/3D/4D GM) and test_copy_shape_dynamic_cube.py (dynamic tail-block copy with 2D/3D/4D GM). 
Each kernel exercises the full nd2nz → dot → fixpipe pipeline.

### DIFF
--- a/src/target/codegen_npuir_api.cc
+++ b/src/target/codegen_npuir_api.cc
@@ -1457,13 +1457,8 @@ void CodeGenTileLangNPUIRAPI::Nz2NdCodegen(const CallNode *op) {
   // Generate hivm.hir.nz2nd for tl.npuir_store_nz2nd.
   tvm::tl::NpuirNz2nd npuirop(op->args, this->vmap);
   // gen memref.subview
-  // src is cc: no min_rank needed.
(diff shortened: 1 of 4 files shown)


CeleNewYear (approved):
LGTM

PR #473, opened 2026-02-12
[AscendNPU-IR][Expert][Developer] support T.IfThenElse in T.Parallel

dazuozcy:
1. support `T.IfThenElse` in `T.Parallel`. This PR only cover this scenario: the condition expressions are vector types (including vector cmps vector, vector cmps scalar).  See `test_ternary_simple` in `vectorization_in_parallel.py` for detail.
2. bugfix: Delete useless local variables in `codegen_npuir_dev.cc`. When `srcShape` is empty, the initialization of the local variable will cause core dump.


dazuozcy:
Has been verified on local env.

CeleNewYear (approved):
LGTM

PR #487, opened 2026-02-13
[AscendNPU-IR] Support wheel package building for TileLang

dangyw18:
Add complete wheel packaging support including CMake integration, backend detection, and TVM bundling. This allows TileLang to be installed via pip install without compilation.

Key features:
- Build wheel package for TileLang
- Bundled TVM with runtime patching
- Automated LLVM dependency management

This setup.py is edited based on the original code from TileLang and keep the original process for CUDA. The default option is USE_NPUIR=true.

### DIFF
--- a/build_wheel.sh
+++ b/build_wheel.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# package_only.sh
+set -e
+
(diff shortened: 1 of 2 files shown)


CeleNewYear (approved):
LGTM

PR #494, opened 2026-02-14
[bug_fix] opt_gdn_chunk_cumsum/o/h/kkt/wy_fast add sync operations in developer mode

BikingNow:
https://www.hiascend.com/document/detail/zh/canncommercial/850/opdevg/BishengCompiler/atlas_bisheng_10_0008.html
ascendc support auto sync in (MTE2 Scalar) (MTE3 Scalar) (Vector Scalar) (Vector Vector)

opt_gdn_chunk_cumsum.py: PIPE_S wait PIPE_V
opt_gdn_chunk_h.py:PIPE_S wait PIPE_V; PIPE_S wait MTE3; pipe_barrier(PIPE_V) when use coeff_ub several times
opt_gdn_chunk_o.py:there are difference in gemm_v0 between ascendc and pto.  MTE1 wait MTE2 before calculation. PIPE_FIX wait PIPE_M after calculation.
opt_gdn_chunk_scaled_dot_kkt.py: pipe_barrier(PIPE_V) when use g_v_ub several times
opt_gdn/opt_gdn_wy_fast.py: pipe_barrier(PIPE_V) when use beta_ub g_ub several times


examples/linear_attention_and_rnn/opt_gdn/opt_gdn_chunk_cumsum.py
```diff
@@ -8,7 +8,7 @@
 '''
 
 pass_configs = {
-	tilelang.PassConfigKey.TL_ASCEND_AUTO_SYNC: False,
+	tilelang.PassConfigKey.TL_ASCEND_AUTO_SYNC: True,
```
benyang0506: Need to investigate why there's an issue with the PTO backend, rather than re-enabling the auto-insert sync pass within the manually inserted sync kernel.

BikingNow: Auto sync will be executed before setvalue/getvalue on ascendc, but we need add sync operations by ourselves on PTO.

PR #459, opened 2026-02-11
[AscendNPU-IR][Expert][Developer]Merge T.copy with T.ascend_copy

happyhappyhappy-arch:

PR #495, opened 2026-02-24
[AscendNPU-IR][Expert][Developer] Enhance slice enforcement for binary vector ops

happyhappyhappy-arch:
This MR require another MR: https://github.com/tile-ai/tilelang-ascend/pull/459

When leading dims is 1, attempt to modify the shape to ensure that the vec op can handle it correctly.

### DIFF
--- a/src/target/codegen_npuir_api.cc
+++ b/src/target/codegen_npuir_api.cc
@@ -175,6 +175,102 @@ getBroadcastDim(const Array<PrimExpr> &buffer_shape0,
   return dims;
 }
 
+// Merge consecutive leading size-1 dimensions into one.
+static Array<PrimExpr> MergeLeadingOnes(const Array<PrimExpr> &shape) {
+  if (shape.empty()) return shape;
+  size_t i = 0;
+  while (i < shape.size()) {
+    const int64_t *v = as_const_int(shape[i]);
+    if (!v || *v != 1) break;
+    i++;
+  }
+  if (i == 0) return shape;
+  Array<PrimExpr> out;
+  out.push_back(IntImm(DataType::Int(32), 1));
+  for (size_t j = i; j < shape.size(); j++) out.push_back(shape[j]);
+  return out;
+}
+
+// Align two shapes to same rank by prepending 1 to the shorter.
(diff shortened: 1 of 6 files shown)


src/target/codegen_npuir_api.cc
```diff
@@ -175,6 +175,102 @@ getBroadcastDim(const Array<PrimExpr> &buffer_shape0,
...
+    }
+    return out;
+  };
+  while (buffer_shape0.size() < buffer_shape1.size()) {
+    buffer_shape0 = prepend_one(buffer_shape0);
+  }
+  while (buffer_shape1.size() < buffer_shape0.size()) {
+    buffer_shape1 = prepend_one(buffer_shape1);
+  }
+}
```
CeleNewYear, line 211: Once we support smart copy, we use rank reduced operation. Why can't we use the same algorithm?

src/target/codegen_npuir_api.cc
```diff
@@ -175,6 +175,102 @@ getBroadcastDim(const Array<PrimExpr> &buffer_shape0,
...
+    int64_t d = result_shape[i];
+    if (mlir::ShapedType::isDynamic(d)) break;
+    stride *= d;
+  }
+  auto layout = mlir::StridedLayoutAttr::get(builder.getContext(), src_offset, strides);
+  auto result_ty = mlir::MemRefType::get(
+      result_shape, ty.getElementType(), layout, ty.getMemorySpace());
+  return builder.create<mlir::memref::CollapseShapeOp>(
+      builder.getUnknownLoc(), result_ty, memref, reassoc);
+}
```
CeleNewYear, line 272: I believe we can reuse the code that handles rank inconsistencies by copying it, instead of implementing two identical copies of the same code.

CeleNewYear (changes requested):

src/target/codegen_npuir_api.cc
happyhappyhappy-arch: This MR has been around for some time now; the initial issue it needed to address was the ad hoc handling of use cases like `T.vadd(A[0, 0, :, :], B[:, :])`. It has fulfilled its historical mission, and I will close this mr later.

PR #497, opened 2026-02-24
[AscendNPU-IR][Expert] Add an example of sparse attention operator with extra optimization and dynamic shape support.

voidLitchi:
Comparing with the original sparse MLA forward kernel:

- Support dynamic shape for `seq_len` and `seq_len_kv`.
- Use `T.macro` to define functions which make the operator more clear and readable.
- Remove unnecessary parameters and loops to optimize the performance.
- Other algorithm adjustments for performance improvement.
- Multi-case accuracy check for dynamic shape verification.

### DIFF
--- a/examples/sparse_mla_fwd_dynamic_shape.py
+++ b/examples/sparse_mla_fwd_dynamic_shape.py
@@ -1,0 +1,533 @@
+# Copyright (c) Huawei Technologies Co., Ltd. 2025.
+
+# Import necessary libraries
+import os
+import glob
+import torch
+
+# Import tilelang modules for NPU kernel development
+import tilelang
+import tilelang.language as T
+
+from typing import Optional
+
+# Clear any cached kernels to ensure fresh compilation
+tilelang.cache.clear_cache()
+
+
+@tilelang.jit(target="npuir")
+def sparse_attn_kernel(
+        batch,
+        heads,
+        dim,
+        tail_dim,
+        top_k,
+        num_kernels,
+        sm_scale=None,
+        block=64,
+):
+    # Validate input parameters
+    assert block % 2 == 0, "Block I size must be even"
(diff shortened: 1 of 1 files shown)


examples/sparse_mla_fwd_dynamic_shape.py
```diff
@@ -0,0 +1,523 @@
...
+        top_k=128,
+        dim=512,
+    )
+
+    generate_and_save_data(
+        case_id=1,
+        batch_size=1,
+        num_heads=32,
+        seq_len=2777,
+        seq_len_kv=4096,
```
CeleNewYear, line 477: Add a group of data to check whether the accuracy is correct when seq_len_kv is 3973.

examples/sparse_mla_fwd_dynamic_shape.py
```diff
@@ -0,0 +1,523 @@
...
+    if errors:
+        error_msg = "\n".join(errors)
+        raise AssertionError(f"Some cases failed:\n{error_msg}")
+    else:
+        print('\033[92mAll checks passed.\033[0m')
+
+
+if __name__ == "__main__":
+    # Specifies which NPU device to use
+    torch.npu.set_device(9)
```
CeleNewYear: The tested server may not have sufficient cards. Change the value to 0.

examples/sparse_mla_fwd_dynamic_shape.py
voidLitchi: Sure, I've already add a test case like that.

examples/sparse_mla_fwd_dynamic_shape.py
voidLitchi: Already fixed.

CeleNewYear (approved):
LGTM

PR #502, opened 2026-02-25
Support auto enable inner-core-pipeline in vector scope while open cross-core-pipeline

erhsh:
Support auto enable inner-core-pipeline in vector scope while open cross-core-pipeline


erhsh:
暂不合入

PR #510, opened 2026-02-27
[WIP][AscendNPU-IR][Developer][skip ci] Support T.alloc_workspace with scope 'global.workspace'

mmyy1217:


happyhappyhappy-arch:
BuildTilelang

happyhappyhappy-arch:
BuildTilelang

src/target/codegen_npuir_dev.cc
```diff
@@ -1716,10 +1913,55 @@ void CodeGenTileLangNPUIRDEV::EmitCopyMemrefToTensor(
...
+            builder.getContext(), mlir::hivm::FixpipePreQuantMode::NO_QUANT);
+        auto pre_relu = mlir::hivm::FixpipePreReluModeAttr::get(
+            builder.getContext(), mlir::hivm::FixpipePreReluMode::NO_RELU);
+        mlir::BoolAttr channel_split = builder.getBoolAttr(false);
+        auto fixpipeOp = builder.create<mlir::hivm::FixpipeOp>(
+            loc, result_tensors, src_tensor, dst_tensor, enable_nz2nd,
+            pre_quant, pre_relu, channel_split);
+        if (is_workspace_dst) {
+          workspace_tensor_map_[npuirop.dst->data.get()] = WorkspaceTensorRecord{
+              fixpipeOp->getResult(0), dstR.offs, dstR.sizes, dstR.strides};
```
AhmedYasser5, line 2026: This might create a faulty MLIR output since nested scopes can create new workspace results that need to be yielded to other scopes

AhmedYasser5 (changes requested):

AhmedYasser5 (approved):
LGTM

src/target/codegen_npuir_dev.cc
```diff
@@ -1716,10 +1978,55 @@ void CodeGenTileLangNPUIRDEV::EmitCopyMemrefToTensor(
...
+      if (use_hivm_fixpipe) {
+        mlir::UnitAttr enable_nz2nd = builder.getUnitAttr();
+        auto pre_quant = mlir::hivm::FixpipePreQuantModeAttr::get(
+            builder.getContext(), mlir::hivm::FixpipePreQuantMode::NO_QUANT);
+        auto pre_relu = mlir::hivm::FixpipePreReluModeAttr::get(
+            builder.getContext(), mlir::hivm::FixpipePreReluMode::NO_RELU);
+        mlir::BoolAttr channel_split = builder.getBoolAttr(false);
+        auto fixpipeOp = builder.create<mlir::hivm::FixpipeOp>(
+            loc, result_tensors, src_tensor, dst_tensor, enable_nz2nd,
+            pre_quant, pre_relu, channel_split);
```
AhmedYasser5, line 2023: I think that using `hivm.hir.fixpipe` with tensors make the result tensors not writable which contradicts with how the workspaces should behave. They should behave just like memory pools where writing data is done in-place (memref style).

src/target/codegen_npuir_dev.cc
```diff
@@ -1625,9 +1713,183 @@ void CodeGenTileLangNPUIRDEV::EmitCopyMemrefToTensor(
...
     mlir::Value src, mlir::Value dst,
     const SliceRange& srcR, const SliceRange& dstR,
-    mlir::Location loc) {
+  mlir::Location loc,
+  bool use_hivm_load) {
   auto dst_tensor_type_ori = dst.getType().cast<mlir::RankedTensorType>();
-  auto src_memref_type_ori = src.getType().cast<mlir::MemRefType>();
+
+  // Workspace GM -> tensor: prefer tensor-form hivm.hir.load so both
+  // ins/outs operands are tensors.
```
AhmedYasser5, line 1729: It would be more straightforward and less error-prone to always use the workspaces in a memref style (in-place modifications). This is feasible because `T.copy` will be always at the start and end of the cube or vector pipelines.

src/target/codegen_npuir_dev.cc
```diff
@@ -1716,10 +1978,55 @@ void CodeGenTileLangNPUIRDEV::EmitCopyMemrefToTensor(
...
+    bool use_hivm_store,
+    bool use_hivm_fixpipe) {
+
+  const bool is_workspace_dst =
+      IsWorkspaceScope(GetPtrStorageScope(npuirop.dst->data));
+
+  auto emit_writeback = [&](mlir::Value src_tensor, mlir::Value dst_memref) {
+    if (use_hivm_fixpipe || use_hivm_store) {
+      mlir::Value dst_tensor = builder.create<mlir::bufferization::ToTensorOp>(
+          loc, dst_memref, /*restrict=*/true, /*writable=*/true);
```
AhmedYasser5, line 2010: For example, getting the corresponding tensor of the original memref can create erroneous MLIR that incorrectly stores the original value into the result.
For example, loading and storing the workspace without changing its loaded value can result in a wrong store:
```py
T.copy(workspace_1, scores)
T.gemm(scores, V_shared, acc_o_c, initC=False)
T.copy(scores, workspace_1)
```

src/target/codegen_npuir_dev.cc
AhmedYasser5, line 4630: This part of the function is another will be also erroneous if `hivm.hir.fixpipe` is used to produce tensors since that part of the function will need to consider yielding only the last tensor of any workspace.
For example, if one workspace is being stored twice in the for loop, the loop will yield the first value not the second.

AhmedYasser5 (changes requested):
Better to use memrefs for storing the values to the workspace

PR #522, opened 2026-02-28
[AscendNPU-IR] Supplement apis.

lzp1021:
1. Add vnot cases.
2. Supports vgather in dev mode and add vgather cases.
3. Fix reshape code mistake.

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -2277,12 +2277,15 @@ void CodeGenTileLangNPUIRDEV::VAtomicAddCodegen(const CallNode *op) {
 
 void CodeGenTileLangNPUIRDEV::VgatherCodegen(const CallNode *op) {
   tvm::tl::NpuirGather npuirop(op->args, this->vmap);
-  Value src = GenSubviewFromRegion(npuirop.src, npuirop.src_range);
-  Value dst = GenSubviewFromRegion(npuirop.dst, npuirop.dst_range);
+  Value src = GetVarValue(npuirop.src);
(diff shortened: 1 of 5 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -2277,12 +2277,15 @@ void CodeGenTileLangNPUIRDEV::VAtomicAddCodegen(const CallNode *op) {
 
 void CodeGenTileLangNPUIRDEV::VgatherCodegen(const CallNode *op) {
   tvm::tl::NpuirGather npuirop(op->args, this->vmap);
-  Value src = GenSubviewFromRegion(npuirop.src, npuirop.src_range);
-  Value dst = GenSubviewFromRegion(npuirop.dst, npuirop.dst_range);
+  Value src = GetVarValue(npuirop.src);
+  Value dst = GetVarValue(npuirop.dst);
```
CeleNewYear, line 2281: Why not use GenExtractSliceFromRegion to support slice gather

testing/npuir/test_vnot.py
```diff
@@ -0,0 +1,91 @@
+import os
```
CeleNewYear, line 1: Change to pytest way

testing/npuir/test_gather.py
```diff
@@ -0,0 +1,102 @@
+import os
```
CeleNewYear, line 1: change to pytest way

CeleNewYear (changes requested):
1. support slice operation and test slice operation
2. use new version of DT (use pytest)

PR #525, opened 2026-02-28
[AscendNPU-IR] Fix out_idx in jit-npu

dangyw18:
This PR fixes several issues in jit_npu.py to properly handle kernels with both tensor and scalar parameters, and ensures correct output indexing regardless of output parameter position. For example, out_idx = [2] where the output is in the middle position of the params instead of the end.
Key Changes:
1. Distinguish between tensor and scalar parameters: scalar param cannot be output.
2. Support arbitrary output parameter positions: out_idx can now point to any parameter index (e.g., out_idx=[2] for middle parameter)

### DIFF
--- a/tilelang/jit/jit_npu.py
+++ b/tilelang/jit/jit_npu.py
@@ -992,28 +992,35 @@ class JitKernel_NPU:
         spec.loader.exec_module(mod)
         self.launch_npu = getattr(mod, "launch")
 
-    def _calcu_grid(self, *args: Any):
+    def _calcu_grid(self, orig_to_input, *args: Any):
+        """
+        Calculate grid dimensions based on symbolic variables and input tensors.
+        
+        Args:
+            tensor_args: List of input tensor arguments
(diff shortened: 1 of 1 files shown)


tilelang/jit/jit_npu.py
```diff
@@ -1218,36 +1197,49 @@ def _extract_param_info(self, func: PrimFunc, out_idx):
...
+
         for i, param in enumerate(params):
             if param in buffer_map:
                 # Tensor parameter (has buffer)
                 buffer = buffer_map[param]
                 dtype_str = str(buffer.dtype)
                 torch_dtype = self._tvm_dtype_to_torch(dtype_str)
                 shape_expr = list(buffer.shape)
-                is_output = (pos_out_idx is not None and i in pos_out_idx)
+
```
dangyw18: do not support scalar as output.
param_info only has 'dtype' and 'shape' attributes.

CeleNewYear (approved):
LGTM

PR #532, opened 2026-03-02
Add an auto infer workspace size function

ac-bcc:
Feature
Add automatic workspace size inference logic for NPU kernel compilation, which enhances the flexibility and adaptability of the compilation process by eliminating the need for manual configuration in most scenarios.

Implementation Details

The core implementation is centered around the new method `_get_workspace_size`, which follows a priority-based strategy to determine the appropriate workspace size for NPU kernels:

1. Environment Variable Priority: If the `TILELANG_ASCEND_WORKSPACE_SIZE` environment variable is set and valid (integer type), its value is used directly as the workspace size. A warning message is printed if the variable is set but not a valid integer, and the logic falls back to auto-inference.

2. Auto-Inference via Kernel Symbol Table: When no valid environment variable is provided, the method parses the symbol table of the compiled kernel library (using `nm -D` command) to find functions ending with the suffix `_infer_workspace_shape_function`. If such functions exist, the library is loaded via `ctypes`, and the function is called to get the accurate workspace size.

3. Default Fallback: If the kernel library does not exist, no target functions are found in the symbol table, or function calls fail, a default workspace size (32768, i.e., 32KB) is used to ensure compilation stability.

The inferred workspace size is then passed to `generate_npu_wrapper_src` to calculate the total workspace size (workspace_size × blockNum) and allocate global memory via `rtMalloc` during kernel launch (only when workspace_size > 0).

Key Advantages

- Reduced Manual Configuration: Developers no longer need to manually set the workspace size in most cases, improving development efficiency.

- Accuracy: By leveraging the kernel's built-in `_infer_workspace_shape_function`, the inferred workspace size matches the actual requirements of the kernel, avoiding waste or insufficiency of memory resources.

- Backward Compatibility: The existing environment variable configuration is retained with the highest priority, ensuring no impact on existing workflows.

- Robustness: Multiple fallback mechanisms (invalid environment variable → auto-inference → default value) ensure the compilation process is not interrupted by unexpected errors.

Notes

The suffix `_infer_workspace_shape_function` for the kernel's workspace inference function is a convention. Ensure that the compiled kernel library includes this function if auto-inference is required.


tilelang/jit/jit_npu.py
```diff
@@ -1493,6 +1530,8 @@ def _npuir_to_bin_enable_npu_compile(self):
             except Exception as e:
                 print(f"error: {str(e)}")
                 sys.exit(1)
+            result = self._get_workspace_size(so_path, "_infer_workspace_shape_function")
```
CeleNewYear, line 1420: An undefined variable was used.

CeleNewYear (changes requested):

tilelang/jit/jit_npu.py
ac-bcc: sry for that, now have solved this problem

CeleNewYear (approved):
LGTM

tilelang/jit/jit_npu.py
```diff
@@ -1029,6 +1030,44 @@ class compiler_npu:
     def __init__(self) -> None:
         pass
 
+    def _get_workspace_size(self, lib_path, suffix, default=32768):
+        #Try to get the infer_workspace_shape_function in the kernel, then use the return value as workspace_size
+        #Use default to avoid except
+        #If you have set the os env "TILELANG_ASCEND_WORKSPACE_SIZE", "TILELANG_ASCEND_WORKSPACE_SIZE" has a higher priority
+        if not os.path.exists(lib64):
```
happyhappyhappy-arch: Is lib64 a var?
Or point to /lib64?
Use shutil.which("nm") instead.

PR #539, opened 2026-03-03
[AscendNPU-IR] Rename npuir interfaces according to naming convention

voidLitchi:
1. Based on the latest naming conventions, npuir has been removed from all interface names, and whether to add v as a prefix is determined based on whether scalar operations and elementwise operations are distinguished.
2. The old interface is still temporarily retained to ensure backward compatibility.
3. The use cases in the `examples` directory have been correctly modified and validated (if they can pass before this submission).


CeleNewYear (approved):
LGTM

PR #541, opened 2026-03-03
[WIP][AscendNPU-IR][Expert] Refact categorized-tests with pytest; support dtype test; support slicing for nd2nz and fixpipe.

mmyy1217:
1. Support T.copy usage for nd2nz and fixpipe
2. Modify test_copy_.._cube testcases to use slice
3. Replace T.Parallel with T.serial in test_copy_shape_dynamic_cube


src/target/codegen_npuir_api.cc
```diff
@@ -1156,10 +1161,75 @@ mlir::Value CodeGenTileLangNPUIRAPI::BinaryOpCodegen(const PrimExprNode *op,
...
+      if (src_dtype == DataType::Float(32) && dst_dtype == DataType::Float(16)) {
+        pre_quant_mode = mlir::hivm::FixpipePreQuantMode::F322F16;
+      } else if (src_dtype == DataType::Float(32) &&
+                 dst_dtype == DataType::BFloat(16)) {
+        pre_quant_mode = mlir::hivm::FixpipePreQuantMode::F322BF16;
+      } else if (src_dtype == DataType::Int(32) &&
+                 dst_dtype == DataType::Int(8)) {
+        pre_quant_mode = mlir::hivm::FixpipePreQuantMode::S322I8;
+      } else {
+        LOG(FATAL) << "Unexpected pre-quant mode in T.copy(L0C, GM).";
```
CeleNewYear, line 1217: For A2 and A3,
If the src is float, the dst can be uint8_t/int8_t/half/bfloat16_t/float as dst
if the src is int32_t, the dst can be uint8_t/int8_t/half/int16_t/int32_t
Cause we need to support enough dtype, Please test this.

CeleNewYear (changes requested):
Please consider the data types we support in detail.

PR #542, opened 2026-03-03
[AscendNPU-IR][Developer] Fix dynamic slice input case for T.copy

happyhappyhappy-arch:
Fix T.copy(BufferLoad, Buffer, size=dynamic_value) where the Buffer side ignored the size parameter and always used its static shape as the region extent.

### DIFF
--- a/tilelang/language/copy.py
+++ b/tilelang/language/copy.py
@@ -144,31 +144,37 @@ def copy(
         extent = list(size)
     else:
         assert src_extent or dst_extent, "Can't deduce copy extents from args"
-        # When one side is BufferLoad (extent=None), use the other side's extent
-        # for block copy.
-        if src_extent is None and dst_extent is not None:
-            extent = list(dst_extent)
-        elif dst_extent is None and src_extent is not None:
-            extent = list(src_extent)
-        else:
-            extent = []
-            for se, de in zip(src_extent, dst_extent):
-                if isinstance(se, tir.IntImm) and isinstance(de, tir.IntImm):
-                    extent.append(se if se.value >= de.value else de)
-                elif isinstance(se, tir.IntImm):
-                    extent.append(de)
-                else:
-                    extent.append(se)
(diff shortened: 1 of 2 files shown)


tilelang/language/copy.py
```diff
@@ -85,7 +85,68 @@ def buffer_region_to_tile_region(buffer_region: tir.BufferRegion, access_type: s
         extents
     ), f"region_extents must be >= extents, region_extents = {region_extents}, extents = {extents}"
 
-    return region(T.BufferLoad(buffer_region.buffer, mins), access_type, *region_extents)
+    result_extents = list(region_extents)
+    n = len(extents)
+    for i in range(n):
+        result_extents[-n + i] = extents[i]
```
CeleNewYear: Why do we need reverse?

tilelang/language/copy.py
```diff
@@ -85,7 +85,68 @@ def buffer_region_to_tile_region(buffer_region: tir.BufferRegion, access_type: s
...
+
+
+def _get_extent(data):
+    """Get extent from data. Buffer→shape, BufferRegion→region extents, else→None."""
+    data = _resolve_var(data)
+    if isinstance(data, tir.Buffer):
+        return list(data.shape)
+    if isinstance(data, tir.BufferRegion):
+        return [x.extent for x in data.region]
+    return None
```
CeleNewYear: Better to keep 
elif isinstance(data, tir.BufferLoad):
            # BufferLoad (e.g. B[bx, by]) cannot infer extent from indices alone.
            # Return None so caller uses the other side's extent for block copy.
            return None

CeleNewYear (commented):
You are advised not to modify unnecessary parts.

tilelang/language/copy.py
happyhappyhappy-arch: Verified. It's not necessary. Removed.

tilelang/language/copy.py
happyhappyhappy-arch: Fixed.

PR #546, opened 2026-03-04
[AscendNPU-IR] Support device-side print only when debug op is defined

shakespearjia:
device-side print only when debug op is defined, no empty infomation for every kernel

### DIFF
--- a/tilelang/jit/jit_npu.py
+++ b/tilelang/jit/jit_npu.py
@@ -264,7 +264,7 @@ def extract_device_print_code_from_cann():
     ])
 
 
-def generate_npu_wrapper_src(constants, signature, workspace_size, mix_mode, lock_num, lock_ini_val):
+def generate_npu_wrapper_src(constants, signature, workspace_size, mix_mode, lock_num, lock_ini_val, need_debug):
     def _ty_to_cpp(ty):
         if ty[0] == "*":
             return "void*"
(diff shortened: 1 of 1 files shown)


tilelang/engine/lower.py
```diff
@@ -239,6 +239,20 @@ def lower(
...
+                found_debug[0] = True
+    
+    for _, func in mod.functions.items():
+        if not isinstance(func, tir.PrimFunc):
+            continue
+        tir.stmt_functor.post_order_visit(func.body, visit)
+
+    if found_debug[0] == True:
+        os.environ['CCE_ENABLE_PRINT'] = 'true'
+
```
CeleNewYear: lower is for optimization, but not for check print op. Please move to jit_npu.

tilelang/jit/jit_npu.py
```diff
@@ -516,7 +516,7 @@ def _format_of(ty):
 #include <vector>
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
-#define __CCE_ENABLE_PRINT__
+{'#define __CCE_ENABLE_PRINT__' if os.environ.get('CCE_ENABLE_PRINT') in ('true', '1') else ''}
```
CeleNewYear: Do not use environment variables for information exchange in internal code.

CeleNewYear (changes requested):

CeleNewYear (approved):
LGTM

PR #564, opened 2026-03-09
[AscendNPU-IR][Developer] The T.parallel feature supports two new scenarios.

voidLitchi:
- Support copy-like and broadcast-like behaviour in T.Parallel
- Support using simple expressions as subscripts of a buffer in T.parallel.
- Add a unit test for this two cases.


CeleNewYear (approved):
LGTM

dazuozcy:
LGTM

PR #569, opened 2026-03-10
[AscendNPU-IR][CI] Simplify npuir pytest-infra logic by allowing only two types of marks.

mmyy1217:
1. Simplify pytest-infra logic by allowing only two types of marks.
2. Update README.md
3. Refact memory_ops and atomic_ops tests.
4. Shrink the test nums of copy_cube.
5. Fix bug that jit_npu cannot find `npu_utils.cpp` by adding fallback


happyhappyhappy-arch:
LGTM

PR #570, opened 2026-03-10
[Add] Add copy_l1_to_l0a/b transpose Function and example.

ArmandAlbert:
Add copy_l1_to_l0a/b transpose Function and example.
The copyshape restriction has been removed; the copy_l1_to_l0ab interface has been refactored, removing layout and adding the transpose option.

### DIFF
--- a/tilelang/language/copy.py
+++ b/tilelang/language/copy.py
@@ -270,7 +270,7 @@ def npu_copy_v2(src: Union[tir.Buffer, tir.BufferLoad, tir.BufferRegion],
     src = _to_region(src, "r")
     dst = _to_region(dst, "w")
 
-    if transpose is not False:
+    if transpose is True:
             return tir.call_intrin("handle", tir.op.Op.get("tl.ascend_copy"), src, dst, enable_relu, transpose)
     else:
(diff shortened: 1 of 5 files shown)


tilelang/language/copy.py
```diff
@@ -267,4 +270,7 @@ def _to_region(data, access_type):
     src = _to_region(src, "r")
     dst = _to_region(dst, "w")
 
-    return tir.call_intrin("handle", tir.op.Op.get("tl.ascend_copy"), src, dst, enable_relu)
+    if transpose is not False:
```
ChaoyangJi: 这里的传参保持数量保持一致吧，传true或者false；

ArmandAlbert: 已修复。

tilelang/language/copy.py
```diff
@@ -267,4 +270,7 @@ def _to_region(data, access_type):
     src = _to_region(src, "r")
     dst = _to_region(dst, "w")
 
-    return tir.call_intrin("handle", tir.op.Op.get("tl.ascend_copy"), src, dst, enable_relu)
+    if transpose is True:
```
lwwangcgz: Don't compare boolean variables with true/false. Just : if transpose, if not transpose...soso

ArmandAlbert: fixed.

PR #579, opened 2026-03-11
[AscendNPU-IR][CI] Multi-threaded acceleration

happyhappyhappy-arch:
Use pytest-xdist for multi-thread acceleration to ensure that each card in the CI can be used.


baberrehman:
Please add a bit of description of the PR.

PR #610, opened 2026-03-16
[AscendNPU-IR][CI] Add compile test before case test

happyhappyhappy-arch:


CeleNewYear:
Why need to add this？

testing/compile/broken/test_atomic_add_1d_dev.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) Tile-AI Corporation.
```
CeleNewYear, line 1: I think we do not need to modify the test directory anymore. There is no need to distinguish between expert mode and developer mode. Our goal is to write code for both expert and developer modes in the future.

CeleNewYear (changes requested):

happyhappyhappy-arch:
> Why need to add this？

When we start adding passes later, we must at least ensure that the previous test cases can still compile successfully. Although there is some overlap with the ops test, the compilation test time is faster, and it can be used as a smoke test to significantly reduce CI time.

testing/compile/broken/test_atomic_add_1d_dev.py
happyhappyhappy-arch: Sure, i will hung up this MR until the mixed mode is implemented.

CeleNewYear:
> > Why need to add this？
> 
> When we start adding passes later, we must at least ensure that the previous test cases can still compile successfully. Although there is some overlap with the ops test, the compilation test time is faster, and it can be used as a smoke test to significantly reduce CI time.

The main bottleneck of the current CI is the installation of the clang environment, test will not take much time

happyhappyhappy-arch:
> > > Why need to add this？
> > 
> > 
> > When we start adding passes later, we must at least ensure that the previous test cases can still compile successfully. Although there is some overlap with the ops test, the compilation test time is faster, and it can be used as a smoke test to significantly reduce CI time.
> 
> The main bottleneck of the current CI is the installation of the clang environment, test will not take much time

The long installation time of clang may be due to other CI programs running simultaneously, competing for resources.Therefore, the time spent in this stage is highly variable, ranging from 1 minute to 30 minutes. One possible improvement is to use g++, which would significantly reduce the size of the installation package.
Can the g++ compile the utils.so also?

CeleNewYear (approved):

PR #615, opened 2026-03-16
[AscendNPU-IR] Fixed the details of the Quick Start document and the Development Guide document

voidLitchi:
1. Fixed the rendering bug in the Development Guide document
2. Add some details in the Quick Start document


docs/快速入门.md
```diff
@@ -49,6 +51,7 @@ git clone https://github.com/tile-ai/tilelang-ascend.git --recursive -b npuir
 cd tilelang-ascend
 bash install_npuir.sh
 source ~/.bashrc
+conda activate tilelang-ascend
```
CeleNewYear: Add an additional note at the end because some users may not use the conda environment.

CeleNewYear (changes requested):

docs/快速入门.md
voidLitchi: Add an annotation and a judgment statement is added for activating Conda env.

CeleNewYear (approved):
LGTM

PR #623, opened 2026-03-17
[AscendNPU-IR] Migrate rms_norm, sparse_mla_bwd, and sparse_mla_fwd operators

lzp1021:
Implement migration for rms_norm, sparse_mla_bwd, and sparse_mla_fwd.


examples/deepseek_v32/sparse_mla_bwd.py
```diff
@@ -0,0 +1,606 @@
+import argparse
```
CeleNewYear: Please use bash format.sh --files your-file-path.
This will make your code conform to clean code standards.

CeleNewYear (changes requested):

PR #627, opened 2026-03-17
[AscendNPU-IR][tvm] Support bfloat16 for binary and unary ops, remove redundant bf16 cast

shakespearjia:
support bfloat16 for binary and unary ops that might use bf16 datatype
remove redundant bf16 cast when multi arith op use the same bf16 BufferRegion


mmyy1217:
Please format the code and check the test failure.

mmyy1217:
LGTM

PR #633, opened 2026-03-18
fix: Replace last step in installation script with a note

AhmedYasser5:
In the installation scripts, it always shows at the end that it added environment updates to `~/.bashrc` and that it sourced it.

This should tell the user that he's ready to use `tilelang` in python, but it's not true because he has to source `~/.bashrc` himself or relaunch a new terminal that will source `~/.bashrc` automatically.

The reason for this is that the user will usually call the script by launching another bash shell instead of sourcing the installation script (not recommended), so sourcing `~/.bashrc` in the installation script will have no effect after the script exits.

### DIFF
--- a/install_npuir.sh
+++ b/install_npuir.sh
@@ -209,14 +209,6 @@ if ! grep -Fq "${BISHENGIR_PY_PKGS}/mlir_core" ~/.bashrc; then
     echo "Added AscendNPU-IR python_packages (mlir_core + bishengir) to PYTHONPATH for NPUIR."
 fi
 
-# Step 12: Source .bashrc to apply changes
-echo "Applying environment changes by sourcing .bashrc..."
-source ~/.bashrc
-if [ $? -ne 0 ]; then
-    echo "Error: Failed to source .bashrc."
-    exit 1
-else
-    echo "Environment configured successfully."
-fi
+echo "NOTE: Please run \"source ~/.bashrc\" or relaunch the terminal to apply the environment changes"
 
 echo "Installation script completed successfully."


install_ascend.sh
```diff
@@ -143,14 +143,6 @@ echo "TileLang path set to: $TILELANG_PATH"
...
-# Step 12: Source .bashrc to apply changes
-echo "Applying environment changes by sourcing .bashrc..."
-source ~/.bashrc
-if [ $? -ne 0 ]; then
-    echo "Error: Failed to source .bashrc."
-    exit 1
-else
-    echo "Environment configured successfully."
-fi
+echo "NOTE: Please run \"source ~/.bashrc\" or relaunch the terminal to apply the environment changes"
```
CeleNewYear: We don't need to change this. Cause this is for ascendC backend.

install_cpu.sh
```diff
@@ -118,15 +118,6 @@ TILELANG_PATH="$(pwd)"
...
-# Step 12: Source .bashrc to apply changes
-echo "Applying environment changes by sourcing .bashrc..."
-source ~/.bashrc
-if [ $? -ne 0 ]; then
-    echo "Error: Failed to source .bashrc."
-    exit 1
-else
-    echo "Environment configured successfully."
-fi
+echo "NOTE: Please run \"source ~/.bashrc\" or relaunch the terminal to apply the environment changes"
```
CeleNewYear: We don't need to change this. Cause this is for CPU backend.

install_cuda.sh
```diff
@@ -151,14 +151,6 @@ echo "Configuring environment variables for TVM..."
...
-# Step 12: Source .bashrc to apply changes
-echo "Applying environment changes by sourcing .bashrc..."
-source ~/.bashrc
-if [ $? -ne 0 ]; then
-    echo "Error: Failed to source .bashrc."
-    exit 1
-else
-    echo "Environment configured successfully."
-fi
+echo "NOTE: Please run \"source ~/.bashrc\" or relaunch the terminal to apply the environment changes"
```
CeleNewYear: We don't need to change this. Cause this is for GPU backend.

install_rocm.sh
```diff
@@ -125,7 +125,6 @@ else
     echo "CUDA_DEVICE_ORDER is already set in ~/.bashrc"
 fi
 
-# Reload ~/.bashrc to apply the changes
-source ~/.bashrc
+echo "NOTE: Please run \"source ~/.bashrc\" or relaunch the terminal to apply the environment changes"
```
CeleNewYear: We don't need to change this. Cause this is for ROCM backend.

CeleNewYear (changes requested):

CeleNewYear (approved):
LGTM

PR #634, opened 2026-03-18
[AscendNPU-IR][Expert] Apply rankreduce in vec-core copy codegen

mmyy1217:
1. Apply rankreduce in vec-core copy codegen


CeleNewYear (approved):
LGTM

PR #640, opened 2026-03-19
[AscendNPU-IR][Dev] Fix the issue with T.flip & T.concat

Ismayil06:
Fixing the issues with T.flip & T.concat and also added the test files.

Changed files: codegen_npuir_dev.cc, +test_vec_concat_dev.py, +test_vec_flip_dev.py

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -2364,13 +2364,17 @@ void CodeGenTileLangNPUIRDEV::VconcatCodegen(const CallNode *op) {
   llvm::SmallVector<Value> srcs;
   size_t n_srcs = npuirop.srcs.size();
   for (size_t i = 0; i < n_srcs; i++) {
-    Value src = GenSubviewFromRegion(npuirop.srcs[i], npuirop.srcs_range[i]);
+    Value src = GenExtractSliceFromRegion(npuirop.srcs[i], npuirop.srcs_range[i]);
     srcs.push_back(src);
   }
   mlir::ValueRange srcs_vr(srcs);
-  Value dst = GenSubviewFromRegion(npuirop.dst, npuirop.dst_range);
-  builder.create<mlir::hivm::VConcatOp>(builder.getUnknownLoc(), TypeRange{},
-                                        dim, srcs_vr, dst);
+  mlir::Value insertBase = NeedGenInsertSlice(npuirop.dst, npuirop.dst_range, srcs[0]);
+  mlir::Type dst_type = insertBase.getType();
+  mlir::TypeRange result_tensors(&dst_type, 1);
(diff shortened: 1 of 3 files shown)


CeleNewYear:
For format-check, you can install ruff in vscode, it will automatically help you format file.

unittest/npuir/test_vec_concat_dev.py
```diff
@@ -1,47 +1,23 @@
-# Tilelang.language.concat
+# Copyright (c) Tile-AI Corporation.
```
CeleNewYear, line 1: Now we have CI, please use pytest method to write test case. And move the test case to testing/npuir

src/target/codegen_npuir_dev.cc
```diff
@@ -2404,10 +2406,14 @@ void CodeGenTileLangNPUIRDEV::VpadCodegen(const CallNode *op) {
 
 void CodeGenTileLangNPUIRDEV::VflipCodegen(const CallNode *op) {
   tvm::tl::NpuirFlip npuirop(op->args, this->vmap);
-  Value src = GenSubviewFromRegion(npuirop.src, npuirop.src_range);
-  Value dst = GenSubviewFromRegion(npuirop.dst, npuirop.dst_range);
-  builder.create<mlir::hivm::VFlipOp>(builder.getUnknownLoc(), TypeRange{}, src,
-                                      dst, npuirop.axis);
+  Value src = GenExtractSliceFromRegion(npuirop.src, npuirop.src_range);
+  Value dst = GetVarValue(npuirop.dst);
```
CeleNewYear: Why here we don't use insertslice method.

src/target/codegen_npuir_dev.cc
```diff
@@ -2366,13 +2364,17 @@ void CodeGenTileLangNPUIRDEV::VconcatCodegen(const CallNode *op) {
...
   for (size_t i = 0; i < n_srcs; i++) {
-    Value src = GenSubviewFromRegion(npuirop.srcs[i], npuirop.srcs_range[i]);
+    Value src = GenExtractSliceFromRegion(npuirop.srcs[i], npuirop.srcs_range[i]);
     srcs.push_back(src);
   }
   mlir::ValueRange srcs_vr(srcs);
-  Value dst = GenSubviewFromRegion(npuirop.dst, npuirop.dst_range);
-  builder.create<mlir::hivm::VConcatOp>(builder.getUnknownLoc(), TypeRange{},
-                                        dim, srcs_vr, dst);
+  Value dst = GetVarValue(npuirop.dst);
```
CeleNewYear: Does the dst do not support slice in AscendNPU IR?

src/target/codegen_npuir_dev.cc
CeleNewYear: ref NeedGenInsertSlice

CeleNewYear (changes requested):

PR #649, opened 2026-03-21
[AscendNPU-IR] Synchronize the community's env

CeleNewYear:
The community has made significant changes to the environment, and we need to maintain consistency with the community to ensure that our subsequent cache implementation is also consistent with the community's.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,10 +54,39 @@ def load_module_from_path(module_name, path):
     return module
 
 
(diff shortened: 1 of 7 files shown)


CeleNewYear:
When I tried cherry-pick, I found that the commit made extensive changes to other parts besides modifying the environment, which prevented cherry-pick from working.

PR #655, opened 2026-03-23
[AscendNPU-IR][Testing] feat: Add mlir tests infra

AhmedYasser5:
This PR provides the ability to add tests to MLIR passes using TileLangIR tools. The tests are added under "testing/mlir" and they automatically locate TileLangIR tools binaries.


CMakeLists.txt
```diff
@@ -69,6 +69,8 @@ if(USE_NPUIR)
     set(Python3_EXECUTABLE "${Python3_EXECUTABLE}" CACHE FILEPATH "Python for tilelangir (from -DPython3_EXECUTABLE)" FORCE)
     message(STATUS "NPUIR/tilelangir Python: ${Python3_EXECUTABLE}")
   endif()
+  include(AddLLVM)
+  include(AddMLIR)
```
happyhappyhappy-arch: Why do we need to include these two targets?

happyhappyhappy-arch:
It looks like we need to add at least FileCheck to CI. Do we need other tools?

AhmedYasser5:
> It looks like we need to add at least FileCheck to CI. Do we need other tools?

I think `FileCheck` should be enough for now. I'll modify the dependencies to remove `not` and `count`.

CMakeLists.txt
AhmedYasser5: Actually, we only need `include(AddLLVM)` because of this error:
image
This function is defined in `AddLLVM.cmake` and it's needed to create test suites.
I'll remove `include(AddMLIR)` until needed.

AhmedYasser5: @happyhappyhappy-arch Do you think it should be added elsewhere?

happyhappyhappy-arch: I think your addition is reasonable.👍

happyhappyhappy-arch:
LGTM

happyhappyhappy-arch (approved):

PR #657, opened 2026-03-23
[AscendNPU-IR][TileLangIR] Add PassPipeline pybind binding and Python wrappers

happyhappyhappy-arch:
The main purpose is to fix the previous issue in pybind that it could not handle non-module nested elements.


AhmedYasser5 (approved):
LGTM

PR #664, opened 2026-03-24
[bugfix] CV sync point quantity unmatch bug fix 

BikingNow:
CV sync point quantity unmatch bug fix

### DIFF
--- a/src/transform/cross_core_pipeline.cc
+++ b/src/transform/cross_core_pipeline.cc
@@ -525,17 +525,16 @@ private:
         std::unordered_set<int> split_indices_V(split_points_V.begin(), split_points_V.end());
         std::unordered_set<const StmtNode*> processed_stmts;
         for (const auto& stmt_info : all_statements_C_) {
-            if (processed_stmts.count(stmt_info.stmt.get())) {
-                continue;
-            }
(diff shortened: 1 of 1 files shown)


src/transform/cross_core_pipeline.cc
```diff
@@ -515,16 +516,19 @@ class LoopRewriter : public StmtMutator {
...
         std::vector<int> split_points_V = ExtractSplitPoints(workspace_writes_V);
-
         std::vector<StageInfo> stages;
         StageInfo current_stage;
         std::unordered_set<int> split_indices_C(split_points_C.begin(), split_points_C.end());
         std::unordered_set<int> split_indices_V(split_points_V.begin(), split_points_V.end());
-
+        std::unordered_set<const StmtNode*> processed_stmts;
         for (const auto& stmt_info : all_statements_C_) {
+            if (processed_stmts.count(stmt_info.stmt.get())) {
```
ArmandAlbert: Redundant continue statements. Maybe modified to `if (!processed_stmts.count(stmt_info.stmt.get())) { ... }` to simplify code could be better.

BikingNow: good idea.done as advised.

PR #668, opened 2026-03-25
[AscendNPU-IR][TileLangIR] feat: Add CVSplit optimization pass

AhmedYasser5:
This pass is grouping ops in each `scf::ForOp` according to loads and stores connected to workspaces. Each group of ops is wrapped into a `scope::ScopeOp` identifying the `hivm::TCoreType` of the scope along with its results. The pass is applied on `scf::ForOp`s in a post-order manner and it looks at the operations from the current nesting level only.

### DIFF
--- a/testing/mlir/Transforms/cv-split.mlir
+++ b/testing/mlir/Transforms/cv-split.mlir
@@ -1,12 +1,191 @@
 // RUN: tilelangir-opt --tilelangir-cv-split %s | FileCheck %s
 
-// CHECK-LABEL: @cv_split
-// CHECK: %[[EMPTY:.*]] = tensor.empty
-// CHECK: %[[RESULT:.*]] = hivm.hir.vmul
-// CHECK-SAME: outs(%[[EMPTY]] :
-// CHECK: return %[[RESULT]] :
-func.func @cv_split(%arg0: tensor<3x3x64xf32>, %arg1: tensor<3x3x64xf32>) -> tensor<3x3x64xf32> {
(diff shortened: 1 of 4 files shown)


testing/mlir/Transforms/cv-split.mlir
```diff
@@ -1,12 +1,191 @@
...
+    %5 = memref_ext.alloc_workspace() : memref<64x64xf32>
+    %6 = memref_ext.alloc_workspace() : memref<64x64xf16>
+    %7 = memref_ext.alloc_workspace() : memref<64x128xf32>
+    %8 = tensor.empty() : tensor<64x128xf16>
+    %9 = tensor.empty() : tensor<64x1xf32>
+    %10 = tensor.empty() : tensor<64x1xf32>
+    %11 = tensor.empty() : tensor<64x128xf32>
+    %12 = tensor.empty() : tensor<64x64xf32>
+    %13 = tensor.empty() : tensor<64x128xf16>
+    %14 = tensor.empty() : tensor<64x128xf32>
```
happyhappyhappy-arch, line 26: The codegen may use the wrong version? For mix mode, in memref level, we only use the Expert version.

happyhappyhappy-arch: Based on the code logic, it seems possible to handle both memref and tensor dialects simultaneously. I think this ut is feasible.

happyhappyhappy-arch (approved):

PR #673, opened 2026-03-25
[Feature] Pipeline workspace splitting, FA expert/auto_sync examples, broadcast & sync fixes

platelett:
## Cross-Core Pipeline
- Rewrite `cross_core_pipeline.cc` with workspace-aware stage splitting
  and exposed-read analysis for automatic buffer double-buffering
- Add workspace-name pairing in `ascend_combinecv.cc`
## Bug Fix
- Fix false dependency in `ascend_sync_insert.cc` and `ascend_memory_planning.cc`:
  intervals that are close but non-overlapping were incorrectly treated as conflicts
- Fix `T.tile.broadcast` crash when dst and src have different ranks
  (e.g. 2D Buffer dst with 3D BufferRegion slice src)
- Restore L1→L0 copy transpose auto-inference: commit a643dbc changed
  the copy_l1_to_l0a/b template signature from layout type (layout::nZ)
  to bool transpose parameter, and commented out print_src_layout in
  AscendCopy::Lower. This broke the link between nZ layout annotation
  and transpose behavior—nZ-annotated buffers silently got transpose=false,
  producing wrong results. Now infer transposeL1 from layout_map in
  AscendCopy::Lower when the source buffer has nZ layout.
## Flash Attention
- Add GQA/MQA support in flash attention examples
  (flash_attn_bhsd_ascendc, auto_pipeline_h16_d128, auto_pipeline_h32_d512)
- Add expert mode FA example (flash_attn_bhsd_expert_h16_d128)
- Add semi-expert mode (auto sync enabled)  FA example (flash_attn_bhsd_auto_sync_h16_d128)
## gemm_v0 API
- Extend `gemm_v0` C tensor validation to match existing A/B pattern:
  accept higher-dimensional C by operating on last-2D slice (batch dims must be 1)
- Update `ascend.py` and `customize.py`


src/op/ascend.cc
```diff
@@ -205,6 +205,14 @@ Stmt AscendCopy::Lower(const LowerArgs &T, arith::Analyzer *analyzer) const {
                << ", dst = " << dst.scope();
   }
 
+  // Auto-infer transpose for L1→L0 copies from layout_map.
```
benyang0506: Revert copy-related changes as they are already handled in a separate PR.
Add `is_transpose` parameter to `copy l1 to l0b` on the frontend to enable transpose support.

**Reference:** [[#570](https://github.com/tile-ai/tilelang-ascend/pull/570)]

benyang0506:
These framework bug fix/enhancement and flash_attn performance optimizations are excellent!
However, many issues stem from the improper placement of the auto-insert CV inter-core synchronization pass. I think automatic inter-core synchronization insertion should be analyzed after inter-core behavior transformations are completed. If this logic were moved to after the passes that separate CV code and inter-core pipeline parallelism, things would be much cleaner. 
This could be considered as a future improvement—worth noting.

PR #684, opened 2026-03-27
update bench_sfa docs

yipyip-c:


benyang0506:
In the bench_sfa README.md, the optimization details for each version could be described more granularly.

PR #696, opened 2026-03-27
[Example] Add RoPE backward kernel with mask-based implementation

liujaaa:
- Implements RoPE forward and backward kernels
- Supports TND [BS, N, D] and BSND [B, S, N, D] input formats
- Supports float16, bfloat16, float dtypes with FP32 accumulation
- Host-side pre-computed mask for gather operation
- Autograd Function wrapper for PyTorch integration


fuhouyu-hw (approved):
/approve

benyang0506 (approved):
lgtm

PR #707, opened 2026-03-28
[AscendNPU-IR][Expert] Fixed bugs in the FA example and modified it to the new recommended usage

voidLitchi:
- Fixed the bug in the example.
- Some old APIs are no longer recommended for use. Update the usage to the latest version.

### DIFF
--- a/examples/flash_attn_npuir.py
+++ b/examples/flash_attn_npuir.py
@@ -1,46 +1,53 @@
 # Copyright (c) Huawei Technologies Co., Ltd. 2025.
-import sys
-import os
 import argparse
 import torch
-
 
 import tilelang
 import tilelang.language as T
 
-torch.npu.set_device(0)
-tilelang.cache.clear_cache()
-
 parser = argparse.ArgumentParser(description="NPU Kernel Compilation")
 
 parser.add_argument(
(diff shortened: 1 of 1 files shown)


CeleNewYear (approved):
LGTM

examples/flash_attn_npuir.py
```diff
@@ -51,104 +58,125 @@ def flashattn(dtype, accum_dtype, seq_len, dim, block_m, block_n, block_k):
...
-            cross_kernel_f32_dim = T.alloc_ub([block_m_half, dim], accum_dtype)
-            cross_kernel_f32_N = T.alloc_ub([block_m_half, block_n], accum_dtype)
-            acc_o = T.alloc_ub([block_m_half, dim], accum_dtype)
-
             acc_c_scale = scale
+            offset_m = cid * block_m
 
             with T.Scope("Cube"):
+                l1_a = T.alloc_L1([block_m, block_share], dtype)
+                l1_b = T.alloc_L1([block_n, block_k], dtype)
```
voidLitchi, line 77: Here is the bug fixed, we used to allocate l1_b with [block_m, block_k], which will cause a runtime error if block_n is larger than block_m

PR #713, opened 2026-03-30
[AscendNPU-IR] Support safe copy in OpenEuler

CeleNewYear:
When we used os.replace, it failed in the OpenEuler system. We changed it so that the fallback on failure would use os.link.


ChangChengShouWang (approved):
LGTM

baberrehman:
LGTM

PR #724, opened 2026-03-31
[AscendNPU-IR] Add MLIR verifier for exp and dev mode

baberrehman:
This PR enables built-in MLIR verifier for dev and expert mode. The verifier captures invalid MLIR errors early and doesn't allow invalid MLIR to be passed to the CANN pipeline.


CeleNewYear (approved):
LGTM

PR #737, opened 2026-03-31
[AscendNPU-IR] Fix runtime for A5

yyyy999:
Fix runtime for A5

### DIFF
--- a/tilelang/jit/jit_npu.py
+++ b/tilelang/jit/jit_npu.py
@@ -242,7 +242,7 @@ def extract_device_print_code_from_cann():
     ])
 
 
-def generate_npu_wrapper_src(constants, signature, workspace_size, mix_mode, lock_num, lock_ini_val, need_debug):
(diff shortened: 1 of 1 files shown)


tilelang/utils/get_ascend_devices.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) Huawei Technologies Co., Ltd. 2025.
+import glob
+import logging
```
CeleNewYear: 1. Don't add this file for getting ascend arch type, there is another file called npu_arch.py
2. Don't need to use smi or pci, we have NPUUtils, there is a func called get_arch

CeleNewYear (changes requested):

tilelang/jit/jit_npu.py
```diff
@@ -1382,12 +1397,22 @@ def _npuir_to_bin_enable_npu_compile(self):
             # Hot fix for CANN 8.5 ends here
 
             npu_compiler_path = get_npucompiler_path()
+            npu_utils = NPUUtils()
+            env_arch = os.getenv("TRITON_ASCEND_ARCH", "")
```
CeleNewYear, line 1401: Why triton?

CeleNewYear (approved):
LGTM

PR #742, opened 2026-04-01
[AscendNPU-IR][Dev] Accuracy fix for tanh op

Ismayil06:
Changed VtanhCodegen function to generate vector hyperbolic tangent approximation using exponential identity in codegen.

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -3115,115 +3115,89 @@ void CodeGenTileLangNPUIRDEV::VerfCodegen(const CallNode *op) {
   }
 }
 
-// Generate vector hyperbolic tangent approximation using polynomial expansion
+// Generate vector hyperbolic tangent approximation using exponential identity
 // in codegen.
 //
 // before(TileLang/TIR semantic):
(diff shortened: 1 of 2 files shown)


CeleNewYear (changes requested):
Please add the an test in testing/npuir, CI will protect vtanh

src/target/codegen_npuir_dev.cc
```diff
@@ -3122,115 +3122,65 @@ void CodeGenTileLangNPUIRDEV::VerfCodegen(const CallNode *op) {
   }
 }
 
-// Generate vector hyperbolic tangent approximation using polynomial expansion
+// Generate vector hyperbolic tangent approximation using exponential identity
```
CeleNewYear, line 3118: For format, you can use bash format.sh --files path-to-your-file

CeleNewYear:
You can also install ruff in your vscode, it will format automatically

src/target/codegen_npuir_dev.cc
Ismayil06: I'm running it right now!

CeleNewYear (approved):
LGTM

PR #743, opened 2026-04-01
[Fix] Fixed the issue of uncertain read order.

ArmandAlbert:
Fixed the issue of uncertain read order.


examples/gemv/example_gemv_v.py
```diff
@@ -5,25 +5,25 @@
 
 import torch
 
+tl.cache.clear_cache()
```
liggest:
```suggestion
```
Please use `tl.disable_cache()` rather than clear all caches
Please add `tl.disable_cache()` under `if __name__ == "__main__":`, so that it won't execute if the operator was imported instead of being directly called

examples/gemv/example_gemv_v.py
```diff
@@ -117,5 +119,6 @@ def main(custom_args=None):
     print("GEMV example passed!")
     print("Kernel Output Match!")
 
+
 if __name__ == "__main__":
-    main()
\ No newline at end of file
+    main()
```
liggest, line 123:
```suggestion
    tl.disable_cache()
    main()
```
Please add `tl.disable_cache()` here

ArmandAlbert: Already fixed.

PR #746, opened 2026-04-01
[AscendNPU-IR][TVM] Support 2D grid for T.kernel both Dev and Exp Mode

shakespearjia:


CeleNewYear (commented):
You can install ruff in vscode, and format on save.

shakespearjia:
> You can install ruff in vscode, and format on save.

I know that, but it's just that the whole files are not formatted, so I want to format the file in another commit

PR #750, opened 2026-04-02
[AscendNPU-IR] Run test when there is a comment

CeleNewYear:
run npuir_test when there is a comment called TilelangFullTest


CeleNewYear:
TilelangFullTest

PR #751, opened 2026-04-02
[AscendNPU-IR][Developer] Fix T.copy gm2ub and ub2gm implicit cast realization

mmyy1217:
## Bug

The current implicit-cast lowering for `T.copy` has two problems:

* **GM -> UB path:** The local UB buffer must use the **source** element type; otherwise, `memref.copy` encounters mismatched element types.
* **Slice-based copy paths:** Attaching the cast to a rank-reduced slice is fragile for downstream passes and backend `vcast` handling, especially when dealing with **dynamic tiles**.

---

## Fix

* **GM -> UB Copies:** Use the **source** element type when creating the local UB buffer.
* **CreateStaticBackedTensor:** Added to build a static-backed tensor for dynamic shapes by combining:
    * Runtime extents from the dynamic source.
    * Static upper bounds from a static-shaped source.
* **InsertSliceWithCast:** * If element types match: Perform a standard `tensor.insert_slice`.
    * If element types differ: First insert into a **shadow tensor** with the source element type, then cast the inserted result to the destination element type.
* **Tensor -> Memref Copies:** Cast the **full source tensor** before creating the rank-reduced `extract_slice`. This ensures the backend `vcast` operates on a full tensor rather than a dynamic rank-reduced slice.

---

## Notes

* This change avoids generating casts on **rank-reduced slice values**.
* For **tensor -> tensor** and **GM -> tensor** paths, moving the cast after `insert_slice` better matches kernel-side behavior (e.g., executing `T.vcast` immediately after a copy).
* For **tensor -> memref**, applying the cast before `extract_slice` prevents the backend from seeing a rank-reduced dynamic value at the cast site.

---

## Tests

* Add implicit-cast `T.copy` coverage for **1D static** cases.
* Add implicit-cast `T.copy` coverage for **1D dynamic tile** cases.

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -1101,6 +1101,44 @@ mlir::Value CodeGenTileLangNPUIRDEV::InsertSlice(
   return insertOp.getResult();
 }
 
+// Helper to handle slice insertion with optional type casting
+mlir::Value CodeGenTileLangNPUIRDEV::InsertSliceWithCast(
+    mlir::Value src_slice, mlir::Value dst, const SliceRange& dstR,
+    mlir::Location loc) {
+  auto srcElemTy = mlir::getElementTypeOrSelf(src_slice.getType());
+  auto dstElemTy = mlir::getElementTypeOrSelf(dst.getType());
+
+  if (srcElemTy == dstElemTy) {
+    return InsertSlice(
+        src_slice, dst,
+        const_cast<llvm::SmallVector<mlir::OpFoldResult> &>(dstR.offs),
+        const_cast<llvm::SmallVector<mlir::OpFoldResult> &>(dstR.sizes),
+        const_cast<llvm::SmallVector<mlir::OpFoldResult> &>(dstR.strides));
+  }
+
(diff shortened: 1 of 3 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -1101,6 +1101,45 @@ mlir::Value CodeGenTileLangNPUIRDEV::InsertSlice(
...
+  auto shadowTy = mlir::RankedTensorType::get(dstTy.getShape(), srcElemTy);
+
+  llvm::SmallVector<mlir::Value> dynamicDims;
+  for (int64_t i = 0; i < dstTy.getRank(); ++i) {
+    if (dstTy.isDynamicDim(i)) {
+      dynamicDims.push_back(builder.create<mlir::tensor::DimOp>(loc, dst, i));
+    }
+  }
+  mlir::Value shadow_empty =
+      builder.create<mlir::tensor::EmptyOp>(loc, shadowTy, dynamicDims);
```
CeleNewYear: When handle tile, dynamicDims will be dynamic, and could not use it to create tensor.empty()

mmyy1217: Fixed

PR #773, opened 2026-04-06
[AscendNPU-IR] Update env ctrl

CeleNewYear:
All newly added environment variables are managed within env.


CeleNewYear (commented):
merge after mix mode and A5 support

PR #779, opened 2026-04-08
[AscendNPU-IR][A5] Support exp and div op for FA

looo000ooong:
Support exp and div op for FA


yyyy999:
Also push up the test for vdiv.

AhmedYasser5 (approved):
LGTM

src/target/codegen_npuir_dev.cc
baberrehman: We do not always generate HIVM op now. Please update the function header to explain the cases when we create unary HIVM op and when we create Linalg op.

baberrehman (changes requested):

baberrehman (approved):

baberrehman:
LGTM

PR #780, opened 2026-04-08
[Add] Support unordered tmp buffer allocate for tmp-remove pass.

ArmandAlbert:
The order of temporary buffer allocation has been modified to support normal allocation of tmp buffer in the case of out-of-order conditions, and it is a prerequisite function for supporting the removal of tmp buffer.

### DIFF
--- a/src/transform/ascend_memory_planning.cc
+++ b/src/transform/ascend_memory_planning.cc
@@ -534,7 +534,12 @@ private:
           alloc_buffer(buffer, current_offset,
                        "Linear memory allocation failed!");
           max_offset = std::max(max_offset, current_offset);
-        } else if (scope == "shared") {
+        }
+      }
+      // Allocate tmp buffer/default buffer after origin buffer to avoid
+      // fragmention in shared memory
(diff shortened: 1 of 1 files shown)


benyang0506 (approved):
Tmp buffer eliminate pass adds allocation for the temporary buffer upon its first use. This PR adapts that behavior.

src/transform/ascend_memory_planning.cc
```diff
@@ -534,7 +534,12 @@ class AscendMemoryPlanning : public arith::IRMutatorWithAnalyzer {
           alloc_buffer(buffer, current_offset,
                        "Linear memory allocation failed!");
           max_offset = std::max(max_offset, current_offset);
-        } else if (scope == "shared") {
+        }
+      }
+      // Allocate tmp buffer/default buffer after origin buffer to avoid
+      // fragmention in shared memory
+      if (scope == "shared") {
```
fuhouyu-hw, line 541: Key strings should be defined using macros

fuhouyu-hw (approved):

PR #682, opened 2026-03-27
[Add] Add default linear memory plan function in MemPlanning Pass.

ArmandAlbert:
Add default linear memory plan function in MemPlanning Pass.

PR #784, opened 2026-04-08
[Fix] Fix gemm rw_mask crash, CrossCorePipeline shared buffer analysis, and FindRelatedBuffers cross-scope overlap

platelett:
## Summary

Fix three independent bugs in the compiler backend.

## Changes

- **Fix gemm C buffer rw_mask**: The accumulator buffer C in `gemm_v0`/`npu_gemm` was unconditionally marked as read-write. When `init=True` (clear accumulator before compute), C is write-only on the first iteration — marking it as read-write causes unnecessary data hazard dependencies in the sync insertion pass. Changed to `"w" if init is True else "rw"` using identity check to also avoid `PrimExpr.__bool__()` crash when `init` is a TVM expression like `k == 0`. Applied to `ascend.py`, `customize.py`, and `pto.py`.

- **Fix shared buffer analysis in CrossCorePipeline**: Removed the `checkBufferScope(buffer) != VEC_SCOPE` filter in `cross_core_pipeline.cc` that incorrectly skipped L1 (shared.dyn scope) buffers during cross-core pipeline analysis, causing L1 buffers shared between Cube and Vector cores to be missed in the pipeline scheduling.

- **Fix scope filtering in FindRelatedBuffers**: `FindRelatedBuffers` in `ascend_sync_insert.cc` only compared physical addresses without checking storage scope. Since memory planning allocates addresses independently per scope, same-core buffers of different scopes (e.g., L1 vs L0C) can share the same address (both starting from 0), causing phantom overlaps and incorrect sync dependencies. Added scope filtering to only match buffers within the same storage scope. Bug surfaced after b36b9508 (#682) made `memory_planning: False` also produce an address_map, but the underlying issue affects both planning modes.


ChaoyangJi (dismissed):

src/transform/ascend_sync_insert.cc
```diff
@@ -35,6 +35,8 @@
 #include "./common/collector.h"
 #include "./common/operation_config.h"
 
+#include "../../3rdparty/tvm/src/tir/transforms/ir_utils.h"
```
LLMZhangYC: Don’t use relative path.

fuhouyu-hw: 不要用相对路径，从tvm开始引入吧

LLMZhangYC (approved):

benyang0506 (approved):

PR #793, opened 2026-04-10
[AscendNPU-IR][Expert] Redesign generic T.copy lowering around aligned views

mmyy1217:
# Expert T.copy Path Rework

This rework replaces the previous ad-hoc behavior of the default NPU generic `T.copy` path with a lowering pipeline built around a single, explicit **shape-and-layout model**.

> **Scope:** The specialized expert copy paths (for example `GM -> L1` and `L0C -> GM`) remain unchanged. This work only restructures the default generic `T.copy` path.

## New Lowering Pipeline

The default path now follows a fixed sequence:

1. **Capture Facts**  
   Capture full-rank slice facts from the source and destination, including:
   - full-rank slice offsets and sizes
   - base contiguous layout strides
   - base linear offset

2. **Align Shape**  
   Derive one logical copy shape by removing only static singleton gaps and matching the remaining backbone dimensions.

3. **Project Layout**  
   Project the real layout from the original contiguous buffers onto the aligned logical shape.
   This preserves the actual base layout instead of inventing a denser one.

4. **Materialize View**  
   Materialize the copy as:
   - a full-rank subview of the original memref region
   - followed by an aligned `memref.reinterpret_cast`
   - and finally a `memref.copy`

---

## Key Improvements

The old generic path mixed:
- region extraction
- singleton reconciliation
- layout construction
- legality decisions

That made it difficult to reason about behavior, especially for:
- rank differences caused by singleton dimensions
- dynamic offsets and extents
- dynamic GM shapes

The new design improves this by:

- **Separating concerns**  
  Extraction, shape alignment, layout projection, and view materialization are now distinct steps.

- **Making the contract explicit**  
  The generic path now has a clear, narrow support contract instead of relying on implicit behavior.

- **Preserving real layout**  
  The projected view keeps the actual contiguous base layout, including dynamic contiguous strides.

- **Supporting dynamic projected strides**  
  For contiguous buffers, projected strides may remain dynamic and are carried through the reinterpret view as runtime values.

---

### Supported
- **Contiguous buffers only**
- **Dynamic range minima**
- **Dynamic range extents**
- **Dynamic contiguous GM shapes**
- **Projected layouts with dynamic strides**, as long as they come from contiguous base layouts
- **Rank mismatch caused only by static singleton gaps**

### Rejected
- Buffers with explicit TIR strides
- Non-contiguous layouts in the generic path
- Shape mismatches that require dropping non-singleton dimensions
- Shape mismatches that require permutation
- Element-type-changing `T.copy`

---

## Front-end Copy Helper

The front-end helper in `tilelang/language/copy.py` has been adjusted to better match the new backend contract while preserving existing tile-style usage.

### Current behavior
- **Whole buffers** remain whole buffers by default.
- **Explicit slices** keep their own extents.
- **BufferLoad operands** still borrow extents from the opposite side when needed.
- **Same-rank mixed slice/plain-buffer copies** are still supported:
  - if one side is an explicit slice
  - and the other side is a plain `Buffer`
  - and both sides have the same rank  
  then the plain `Buffer` side borrows the explicit slice extent so tile-style copies like
  `T.copy(A[bx:..., by:...], UB)` continue to materialize a matching logical tile.

This keeps common tail-tile copy patterns working without changing the specialized backend paths.

---

## Testing

The generic copy tests have been rewritten to exercise the support boundary directly.

They cover:
- exact-shape copies
- singleton insertion/removal
- dynamic `min`
- dynamic `extent`
- dynamic GM shape cases
- dynamic projected stride cases
- plain-UB tail-tile copies

The unsupported cases are kept separate so the contract remains explicit.

---

## Summary

This rework makes the default generic `T.copy` path:

- more predictable
- easier to reason about
- more explicit about legality
- more faithful to the real contiguous buffer layout

The result is a generic path that is structured enough to defend, while still supporting important dynamic contiguous-copy cases that appear in real kernels.


src/target/codegen_npuir_api.cc
```diff
@@ -1164,6 +1030,320 @@ mlir::Value CodeGenTileLangNPUIRAPI::BinaryOpCodegen(const PrimExprNode *op,
...
+  llvm::SmallVector<mlir::OpFoldResult> strideOfrs;
+  strideOfrs.reserve(layout.viewStrides.size());
+  for (int64_t stride : layout.viewStrides) {
+    strideOfrs.push_back(builder.getIndexAttr(stride));
+  }
+
+  return builder.create<mlir::memref::ReinterpretCastOp>(
+      builder.getUnknownLoc(), targetTy, fullSubview, layout.viewOffset,
+      layout.viewSizes, strideOfrs);
+}
```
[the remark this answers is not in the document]
mmyy1217: /gemini 
I don’t think this is a bug.
ComputeProjectedViewOffset intentionally computes an offset relative to the base allocation, and that is still correct when BuildAlignedCopyView reinterprets a SubViewOp. In MLIR, memref.reinterpret_cast uses the source’s underlying memory, not the source view’s offset as an additive base. That is different from memref.subview, which is view-relative.
So applying layout.viewOffset to fullSubview does not double-count the subview offset.
BuildFullRankSubview may still be simplifiable as a separate cleanup, but I would not change this path as an offset fix.

src/target/codegen_npuir_api.cc
```diff
@@ -1238,11 +1418,46 @@ void CodeGenTileLangNPUIRAPI::AscendCopyCodegen(const CallNode *op) {
...
+  mlir::Value srcFullSubview = BuildFullRankSubview(srcFacts);
+  mlir::Value dstFullSubview = BuildFullRankSubview(dstFacts);
+  mlir::Value srcAlignedView = BuildAlignedCopyView(
+      srcFullSubview, srcFacts.baseMemrefType.getElementType(),
+      srcFacts.baseMemrefType.getMemorySpace(), shapePlan.alignedShape,
+      srcLayout);
+  mlir::Value dstAlignedView = BuildAlignedCopyView(
+      dstFullSubview, dstFacts.baseMemrefType.getElementType(),
+      dstFacts.baseMemrefType.getMemorySpace(), shapePlan.alignedShape,
+      dstLayout);
```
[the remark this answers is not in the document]
mmyy1217: /gemini
I don’t think switching this to baseMemref is necessary as a correctness fix.
Using baseMemref instead may be a valid refactor direction if we want to simplify the generic path, but it is not required to preserve correct offset semantics. The current BuildFullRankSubview -> BuildAlignedCopyView split is still semantically valid.

src/target/codegen_npuir_api.cc
```diff
@@ -1164,6 +1030,320 @@ mlir::Value CodeGenTileLangNPUIRAPI::BinaryOpCodegen(const PrimExprNode *op,
...
+  if (plan.alignedShape.empty()) {
+    // Scalar-like copies still need one representative singleton dimension so
+    // the generic path can materialize a legal memref.copy view.
+    ICHECK(!srcInfo.gapIdx.empty() && !srcInfo.gapIdx.front().empty() &&
+           !dstInfo.gapIdx.empty() && !dstInfo.gapIdx.front().empty())
+        << "generic T.copy could not derive a valid target shape.";
+    plan.alignedShape.push_back(1);
+    plan.srcKeptDims.push_back(srcInfo.gapIdx.front().front());
+    plan.dstKeptDims.push_back(dstInfo.gapIdx.front().front());
+  }
```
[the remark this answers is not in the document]
mmyy1217: /gemini
I don’t think this is a practical issue for the current contract.
This path is not expected to see true rank-0 buffers in our current TileLang/NPUIR pipeline, so I would not treat this as a blocking fix for this patch.
If we decide to support rank-0/scalar memrefs later, I would rather handle that explicitly as a separate scalar-copy case than add a fallback here.

src/target/codegen_npuir_api.cc
```diff
@@ -522,61 +341,87 @@ CodeGenTileLangNPUIRAPI::GetHIVMAddressSpace(String address_space) {
...
+                               MaterializeFn &&materialize) {
+  llvm::SmallVector<mlir::OpFoldResult> ofrs;
+  ofrs.reserve(exprs.size());
+  arith::Analyzer analyzer;
+  for (PrimExpr expr : exprs) {
+    PrimExpr simplified = analyzer.Simplify(expr);
+    if (auto intImm = as_const_int(simplified)) {
+      ofrs.push_back(builder.getIndexAttr(*intImm));
+    } else {
+      ofrs.push_back(materialize(simplified));
```
CeleNewYear: This is a common function that appears multiple times in this PR. It can be extracted and optimized into a common function.

src/target/codegen_npuir_api.cc
```diff
@@ -522,61 +341,87 @@ CodeGenTileLangNPUIRAPI::GetHIVMAddressSpace(String address_space) {
...
+  llvm::SmallVector<int64_t> staticValues;
+  staticValues.reserve(exprs.size());
+  arith::Analyzer analyzer;
+  for (PrimExpr expr : exprs) {
+    PrimExpr simplified = analyzer.Simplify(expr);
+    if (auto intImm = as_const_int(simplified)) {
+      staticValues.push_back(*intImm);
     } else {
-      shape.emplace_back(ShapedType::kDynamic);
-      isDynamicShape = true;
```
CeleNewYear, line 557: This is a common function that appears multiple times in this PR. It can be extracted and optimized into a common function.

tilelang/language/copy.py
```diff
@@ -131,55 +136,74 @@ def _is_slice(data):
...
             data = T.get_let_value(data)
         if isinstance(data, tir.Buffer):
+            if not has_explicit_size:
+                # When a plain buffer is paired with an explicit slice of the
+                # same rank, reuse the peer extents so tail-tile copies like
+                # T.copy(A[bx:..., by:...], UB) and T.copy(UB, C[bx:..., by:...])
+                # keep matching logical shapes. For rank-mismatch singleton
+                # cases, preserve whole-buffer semantics and let the backend
+                # perform shape alignment.
+                if (
```
CeleNewYear, line 174: Please refer to the newest version of the Tilelang implementation to see if this method of completely preserving the shape conflicts with automatic tail block handling.

CeleNewYear (changes requested):

src/target/codegen_npuir_api.cc
mmyy1217: Fixed

src/target/codegen_npuir_api.cc
mmyy1217: Fixed

CeleNewYear (commented):
Please rebase

tilelang/language/copy.py
mmyy1217: It doesn't conflict with automatic tail block handling. The method here leaves tail-handling to the codegen stage to avoid adding complex shape-alignment logic in copy.py. Auto-tail handling is not supported currently. We suggest users write logical-shape aligned copy now to avoid triggering this path.

CeleNewYear (approved):
LGTM

PR #808, opened 2026-04-13
[AscendNPU IR][A5] fix: use more friendly arith+math ops instead of linalg

AhmedYasser5:
Ops like `linalg.add` are not handled correctly in NPU IR, so we change to `arith` and `math` dialects to support complex A5 kernels like FA.

### DIFF
--- a/src/target/codegen_npuir_dev.cc
+++ b/src/target/codegen_npuir_dev.cc
@@ -1953,8 +1953,8 @@ static inline constexpr bool startsWith(std::string_view str,
 /// before:
 ///     T.npuir_exp(A, B)
 /// after:
-        %.* = <linalg>.<op> ins(%A_trans) outs(B) -> tensor<>
-        or
+///        %.* = <linalg>.<op> ins(%A_trans) outs(B) -> tensor<>
+///        or
 ///     %.* = hivm.hir.<op> ins(A) outs(B) -> tensor<>
 template <typename T, typename U>
 void CodeGenTileLangNPUIRDEV::UnaryVecOpCodegen(const CallNode *op) {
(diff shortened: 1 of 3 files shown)


src/target/codegen_npuir_dev.cc
```diff
@@ -3439,9 +3450,24 @@ mlir::Value CodeGenTileLangNPUIRDEV::VisitExpr_(const CallNode *op) {
...
+      const auto elementType =
+          dst.getType().cast<ShapedType>().getElementType();
+      if (elementType.isa<IntegerType>())
+        return builder
+            .create<mlir::arith::AddIOp>(builder.getUnknownLoc(), src0, src1)
+            .getResult();
+      return builder
+          .create<mlir::arith::AddFOp>(builder.getUnknownLoc(), src0, src1)
+          .getResult();
+    });
```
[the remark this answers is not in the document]
AhmedYasser5: maybe no need because the function is there to make it as general as possible

CeleNewYear (commented):
These changes are excellent in A5, but after being merged into the mainline, the test cases will run CI in the mainline A2/A3, so the compilation options need to be appropriately supported and modified.

AhmedYasser5:
> These changes are excellent in A5, but after being merged into the mainline, the test cases will run CI in the mainline A2/A3, so the compilation options need to be appropriately supported and modified.

I've created an issue and we shall track this later before merging with `npuir` branch

src/target/codegen_npuir_dev.cc
```diff
@@ -3471,13 +3496,66 @@ mlir::Value CodeGenTileLangNPUIRDEV::VisitExpr_(const CallNode *op) {
   } else if (op->op.same_as(Op::Get("tl.npuir_bitcast"))) {
     BitcastCodegen(op);
   } else if (op->op.same_as(Op::Get("tl.npuir_div"))) {
-    CreateHIVMBinaryVectorOp<mlir::hivm::VDivOp>(op);
+    CreateHIVMBinaryVectorOp(op, [](OpBuilder &builder, Value src0, Value src1,
```
baberrehman: We need to abstract this logic inside CreateHIVMBinaryVectorOp (as in A2/A3) lowering and also handle scalar case.

Also, why do we need to pass `[](OpBuilder &builder, Value src0, Value src1, Value dst` as arguments? They can be derived from op.

baberrehman (changes requested):

src/target/codegen_npuir_dev.cc
AhmedYasser5: for some operations, `FloatType` should be handled differently from `IntegerType`. And sometimes, `unsigned int` is handled differently from `signed/signless int`. However, let me try to do it in the function instead.

baberrehman (approved):
LGTM

PR #811, opened 2026-04-14
[AscendNPU-IR][MIX]Use graph-solver replace inject-sync

happyhappyhappy-arch:


tilelang/tladapter/transforms/bishengir.py
```diff
@@ -7,4 +7,5 @@
 adapt_triton_kernel = pass_fn("adapt-triton-kernel")
 bind_workspace_arg = pass_fn("hivm-bind-workspace-arg", "func.func")
 infer_workspace_size_func = pass_fn("hivm-insert-infer-workspace-size-func", "func.func")
+graph_sync_solver = pass_fn("hivm-graph-sync-solver", "func.func")
```
CeleNewYear, line 10: Because we've locked the 3rdparty commit, we need to ensure that the pass was ready before we locked the commit before making this change.

CeleNewYear (changes requested):

PR #822, opened 2026-04-15
[AscendNPU-IR]Hot fix for the same name kernel conflict in ci

happyhappyhappy-arch:


CeleNewYear (approved):
LGTM

CeleNewYear (commented):
A more complete and reliable solution would be to have the cache maintained uniformly, placing the launcher and the kernel's own cache at the same address. However, can be done later.

PR #843, opened 2026-04-16
[AscendNPU-IR] Support CANN-9.0.0

happyhappyhappy-arch:
This MR will be suspended until CANN releases version 9.0.0.

### DIFF
--- a/install_npuir.sh
+++ b/install_npuir.sh
@@ -45,11 +45,10 @@ normalize_cann_version() {
 infer_cann_version_from_env() {
     local current_home="${ASCEND_HOME_PATH}"
     if [ -z "$current_home" ]; then
-        echo "Error: neither --cann-version nor ASCEND_HOME_PATH is set." >&2
-        echo "Please choose one of the following before re-running:" >&2
-        echo "  1. Pass --cann-version=<${SUPPORTED_CANN_VERSIONS[*]}>" >&2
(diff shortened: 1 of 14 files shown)


install_npuir.sh
```diff
@@ -13,10 +13,67 @@ export PATH="${PYTHON_DIR}:$PATH"
...
+        *)
+            return 1
+            ;;
+    esac
+}
+
+infer_cann_version_from_env() {
+    local current_home="${ASCEND_HOME_PATH}"
+    if [ -z "$current_home" ]; then
+        echo "Error: neither --cann-version nor ASCEND_HOME_PATH is set." >&2
```
CeleNewYear: Please use version 8.5.0 by default, as it is the version currently recommended by the community.

pyproject.toml
```diff
@@ -13,7 +13,7 @@ column_limit = 100
 indent_width = 4
 
 [tool.codespell]
-ignore-words-list = "nd, te, ist, LOD, offen, NotIn, HSA, cann"
+ignore-words-list = "nd, te, ist, LOD, offen, NotIn, HSA, cann, VOr"
```
CeleNewYear, line 16: What VOr mean?

happyhappyhappy-arch: The Vector Or opration, which in README.md as VOr.
In format CI, that op considered as a misspelling of "for".

install_npuir.sh
happyhappyhappy-arch: Checked

PR #851, opened 2026-04-17
[AscendNPU-IR][Expert] Fix FA reduce block_n_tail bug

Wamdus3:
Fixed FA reduce block_n_tail bug


examples/flash_attn_npuir.py
```diff
@@ -207,7 +207,7 @@ def FlashAttnExp(
 
                     T.vmul(cross_kernel_f32_N, acc_c_scale, cross_kernel_f32_N)
                     T.reduce(
-                        cross_kernel_f32_N, scores_max, dims=[1], reduce_mode="max"
+                        cross_kernel_f32_N, scores_max, dims=[1], reduce_mode="max", size=[real_m, tail_size_n]
```
CeleNewYear, line 210: We previously removed `size` because `slice` syntax was more precise; why is `size` being reinstated here?

CeleNewYear (changes requested):

PR #858, opened 2026-04-17
[AscendNPU-IR][A5] alloc test cases

baberrehman:
Test case for NPU memory allocation ops in Developer mode:
  - T.alloc_var   (local.var scope)
  - T.alloc_L0A   (wmma.matrix_a scope)
  - T.alloc_L0B   (wmma.matrix_b scope)
  - T.alloc_L0C   (wmma.accumulator scope)
  - T.alloc_L1    (shared.dyn scope)
  - T.alloc_ub    (shared / UB scope)

The kernel allocates buffers in every memory scope, performs a simple
computation that touches each buffer, and writes the result to global memory
so we can verify correctness.

### DIFF
--- a/testing/npuir/memory_ops/test_alloc_dev.py
+++ b/testing/npuir/memory_ops/test_alloc_dev.py
@@ -1,0 +1,99 @@
+# Copyright (c) Tile-AI Corporation.
+# Licensed under the MIT License.
+"""
+Test case for NPU memory allocation ops in Developer mode:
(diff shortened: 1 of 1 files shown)


CeleNewYear (commented):
In developer mode, we generally use a simplified memory mapping relationship, using alloc_shared to correspond to T.alloc_L1 and alloc_ub; and alloc_fragment to correspond to alloc_L0A, alloc_L0B, alloc_L0C, and alloc_var.

PR #873, opened 2026-04-20
[AscendNPU IR][A5] feat: support transpose, gather, print

AhmedYasser5:
This pull request updates the NPUIR developer mode codegen to use the hfusion dialect for gather, transpose, and print operations, and adds corresponding unit tests. Feedback identifies a runtime casting error in VgatherCodegen where mlir::RankedTensorType should be replaced with mlir::ShapedType to avoid failures with MemRef types, and a bug in the gather test script due to an undefined variable itype.


baberrehman:
LGTM

PR #879, opened 2026-04-20
[AscendNPU-IR][Expert] Bug fix for the codegen of binary vector operators

voidLitchi:
- The codegen of binary operators is reconstructed.
- The operands are no longer required to have the same buffer shape, but only to have the same region shape.

### DIFF
--- a/src/op/ascend.cc
+++ b/src/op/ascend.cc
@@ -21,6 +21,23 @@ namespace tl {
 
 using namespace tir;
 
+NpuirOperand NpuirOperand::FromExpr(const PrimExpr &expr,
+                                    const BufferMap &vmap) {
+  if (const auto *call = expr.as<CallNode>()) {
+    // CallNode should be a tensor
+    auto region = RegionOp(call->args, vmap);
+    return NpuirOperand::Tensor(region.GetBuffer(), region.GetRanges());
+  }
+  if (expr.as<IntImm>() || expr.as<FloatImm>() || expr.as<tir::VarNode>()) {
+    // If there are other types of nodes that need to be treated as scalars,
+    // please add them here.
+    return NpuirOperand::Scalar(expr);
+  }
+  LOG(FATAL) << "NpuirOperand::FromExpr cannot handle the expr with type of \""
+             << expr->GetTypeKey() << '\"';
+  __builtin_unreachable();
+}
+
 AscendCopy::AscendCopy(Array<PrimExpr> args, BufferMap vmap) : args_(args) {
   Array<Range> rgs[2];
   Buffer bf[2];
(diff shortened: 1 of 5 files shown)


src/op/ascend.cc
```diff
@@ -21,6 +21,22 @@ namespace tl {
 
 using namespace tir;
 
+NpuirOperand NpuirOperand::FromExpr(const PrimExpr &expr,
+                                    const BufferMap &vmap) {
+  if (const auto *call = expr.as<CallNode>()) {
+    // CallNode should be a tensor
+    auto region = RegionOp(call->args, vmap);
```
CeleNewYear, line 28: check std::tie

src/op/ascend.h
```diff
@@ -19,6 +19,59 @@ namespace tl {
 
 using namespace tir;
 
+class NpuirOperand {
+public:
+  enum class Kind {
+    kTensor,
+    kScalar,
+  };
```
CeleNewYear: Private or Protect is a better way

src/target/codegen_npuir_api.cc
```diff
@@ -788,6 +788,19 @@ static bool IsStaticIntOFR(mlir::OpFoldResult ofr, int64_t value) {
...
+bool CodeGenTileLangNPUIRAPI::EnableToReduceRank(Array<Range> range,
+                                                 int target_rank) {
+  int min_rank = range.size();
+  for (Range r : range) {
+    if (analyzer_->CanProveEqual(r->extent, 1)) {
+      --min_rank;
+    }
+  }
+  return min_rank <= target_rank;
+}
```
CeleNewYear: There is a exsiting algorithm, please talk with zuoming

src/op/ascend.cc
voidLitchi: Since `NpuirOperand::FromExpr` returns a constructor, which is an rvalue, it is not possible to directly use `std::tie`, which requires an lvalue, here. Converting an rvalue to an lvalue in this context would only incur additional overhead, so it is not recommended to use std::tie here.

src/op/ascend.h
voidLitchi: Thank you, you are right. Here I turned to define it as a private enum class.

src/target/codegen_npuir_api.cc
voidLitchi: I discussed with the author of the `GenRankReducedSubviewFromRegion` method and concluded that the definition of this function indeed lacks rigor. The current solution is to temporarily remove it. 
On one hand, even if this check is not performed here, an error related to shape inconsistency will still be reported when the `getBroadcastDim` method is executed later. 
On the other hand, we can consider improving the checks in `GenRankReducedSubviewFromRegion` in the future, and once the relevant refactoring is completed, the shape check can be re-added here.

CeleNewYear (approved):
LGTM

src/target/codegen_npuir_api.cc
```diff
@@ -1888,25 +1888,17 @@ void CodeGenTileLangNPUIRAPI::BitcastCodegen(const CallNode *op) {
...
-  auto buffer = op->buffer;
-  auto indices = op->indices;
-
-  // Check pre-conditions
-  if (op->dtype.lanes() != 1) {
-    LOG(FATAL) << "lanes not one";
-  }
-  if (op->dtype != buffer->dtype) {
-    LOG(FATAL) << "The load type and buffer element type do not match";
-  }
```
yyyy999, line 1929: Removing the dtype consistency check means that if the region's dtype does not match the buffer dtype, an incorrect MLIR load instruction will be silently generated, potentially causing runtime data corruption .

src/op/ascend.cc
```diff
@@ -21,6 +21,22 @@ namespace tl {
...
+                                    const BufferMap &vmap) {
+  if (const auto *call = expr.as<CallNode>()) {
+    // CallNode should be a tensor
+    auto region = RegionOp(call->args, vmap);
+    return NpuirOperand::Tensor(region.GetBuffer(), region.GetRanges());
+  }
+  if (expr.as<IntImm>() || expr.as<FloatImm>() || expr.as<tir::VarNode>()) {
+    // If there are other types of nodes that need to be treated as scalars,
+    // please add them here.
+    return NpuirOperand::Scalar(expr);
```
yyyy999, line 34: when allows args[1] to be a scalar ( IntImm / FloatImm / VarNode ), then src1_call == nullptr , it will causing BF16 legalization to be silently skipped . If src0 is a BF16 tensor and src1 is a scalar, the operator will execute without BF16leagalize, potentially leading to precision errors or unsupported hardware operations .

src/target/codegen_npuir_api.cc
```diff
@@ -1917,59 +1909,98 @@ CodeGenTileLangNPUIRAPI::GenMemrefLoadFromRegion(const BufferLoadNode *op) {
...
+  ICHECK(binaryOp.Src0().IsTensor() || binaryOp.Src1().IsTensor())
+      << "Binary Vector Ops does not support cases where both input operands "
+         "are scalars.";
+  // src_dtype is used to unify the type of input operands.
+  DataType src_dtype;
+  if (binaryOp.Src0().IsTensor()) {
+    src_dtype = binaryOp.Src0().GetBuffer()->dtype;
+  } else {
+    src_dtype = binaryOp.Src1().GetBuffer()->dtype;
+  }
```
yyyy999, line 1949: Suggest when both operands are tensors, add ICHECK(src0_dtype == src1_dtype) or at least log a warning.

src/op/ascend.cc
```diff
@@ -21,6 +21,22 @@ namespace tl {
...
+  }
+  LOG(FATAL) << "The type of expr cannot be handled by NpuirOperand.";
```
yyyy999: Suggest to include more detailed information in the logs. eg: 
LOG(FATAL) << "NpuirOperand::FromExpr: unsupported expr type: " << expr->GetTypeKey();

src/op/ascend.cc
voidLitchi: I noticed that BF16leagalize is a TVM Pass, and TVM Pass should run before the codegen. So in our pipeline, the code change in codegen could not influence the behavior of a TVM Pass like BF16leagalize.

src/target/codegen_npuir_api.cc
voidLitchi: Well, actually, if both operands are tensors, the variable `src_dtype` won't be used, so it might be fine. But of course, this is a bit unfriendly for maintenance, and it's a good idea to add a check for the data type between the two tensors here. Thanks for your suggestion, I'll make the changes.

src/op/ascend.cc
voidLitchi: Good idea, thank you.

src/target/codegen_npuir_api.cc
voidLitchi: Since I changed the arguments of this method, we can no longer get the op->dtype here. However, due to the limitations of the tilelang DSL interface, we will not actually encounter a scenario where op->dtype and op->buffer->dtype are different in this case. On the other hand, I have already validated the data types before calling this function. Therefore, in summary, this validation can be removed here.

PR #901, opened 2026-04-21
test: re-add xfail for float16 reduction sum

ShareableXue:
Float16 has a rare probabilistic overflow issue when summing large
values, causing inf or precision errors. Removing xfail made the test
flaky. Restore xfail to avoid random CI failures.


fuhouyu-hw (approved):

benyang0506 (approved):
/approve

PR #917, opened 2026-04-24
[Add] Add DeepSeek V4 kernels and document.

ArmandAlbert:
Add DeepSeek V4 kernels and document.


fuhouyu-hw (dismissed):

examples/deepseek_v4/act_quant.py
```diff
@@ -0,0 +1,186 @@
+# coding=utf-8
+# This program is free software, you can redistribute it and/or modify it.
```
benyang0506: Please confirm whether the license here should use CANN or MIT, keeping it consistent with Tile AI.

ArmandAlbert: License has been changed to Tile-AI (MIT).

benyang0506 (approved):
/approve

fuhouyu-hw (approved):

PR #934, opened 2026-04-27
[bugfix][sfa] fix test input generation to use true random top-k indices

benyang0506:
Fix test input generation to use true random top-k indices.

- replace the dense scatter-mask golden path with a gather-based sparse reference
- preserve duplicate top-k indices and support -1 masked entries
- fix test input generation to use true random top-k indices


fuhouyu-hw (approved):

LLMZhangYC (approved):
approve

PR #942, opened 2026-04-28
[AscendNPU-IR] 1. Support T.Cast in Parallel 2. Enhance the test case…

endaiHW:


src/transform/npu_loop_vectorize.cc
```diff
@@ -357,6 +364,10 @@ class LoopDecompose : public StmtMutator {
   // Unary: region_in -> out
   Stmt BuildUnaryStmt(const std::string &op_name, const PrimExpr &region_in,
                       const BufferAccessInfo &out) {
+    if (op_name == "Cast") {
+      return BuildNpuirCall(
+          op_name, {region_in, BuildRegionCall(out, 2, 1), StringImm("rint")});
```
yyyy999, line 374: The round mode is hardcoded as "rint" , for integer-to-float or float-to-lower-precision casts where floor or trunc semantics are required, this will produce silent numerical errors .

Either (a) extend CastNode handling to carry round mode metadata, or (b) document this limitation clearly and add a warning when Cast is encountered in a Parallel context with a non-default round mode expectation. At minimum, add a comment explaining the hardcoded "rint" choice.

PR #984, opened 2026-05-08
[Fix] Align pytest skip ignores 🤖

lwwangcgz:
## Summary

Align the pytest skip check with the existing changed-file ignore rules.

## Changes

- Ignore `.agents/` files when deciding whether `bench_test.sh` can use `--skip-pytest`.
- Ignore `.md` and `.png` files in the same skip check.

## Root Cause

PR #957 added `.agents/` and docs/image ignores to workflow triggers and `check_changes`, but `Check if skip pytest` had a separate `git diff` loop that still treated `.agents/` files as outside `examples/`, which forced `skip_pytest=false`.

## Testing

- `git diff --check upstream/ascendc_pto...HEAD`

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -392,6 +392,14 @@ jobs:
           skip_pytest=true
           
           for file in $changed_files; do
+            if [[ "$file" =~ ^\.agents/ ]]; then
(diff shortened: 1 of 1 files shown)


ChaoyangJi (approved):
lgtm

LLMZhangYC (approved):
approve

PR #998, opened 2026-05-11
[skill] add skill-journal & tilelang-skill-review feedback loop

hedi515:
- skill-journal: schema for collecting per-op skill feedback entries
- tilelang-skill-review: aggregates pending entries and applies approved changes
- op-generate: add section 6 to emit feedback into skill-journal after each run


benyang0506 (dismissed):
/lgtm

PR #1008, opened 2026-05-12
[Feature] Add GATHER support for PTO

hedi515:


hedi515:
test

src/target/codegen_ascend_pto.cc
```diff
@@ -1447,6 +1449,65 @@ void CodeGenTileLangAscendPto::GatherbCodegen(const CallNode *op,
...
+                             "elem_size, got: "
+                          << x;
+      x >>= 1;
+      ++r;
+    }
+    return r;
+  };
+  int shift_div = log2_int(elem_size);
+
+  // mask /= elem_size  (byte offset -> element offset; TGATHER reads indices
```
ShareableXue, line 1498: Great work! The implementation cleverly accepts byte offsets from the frontend and automatically inserts a right shift in the PTO backend codegen to convert them into element indices, seamlessly adapting to the PTO interface.

That said, this conversion does introduce some performance overhead. Would it be possible to document this behavior in our docs/skills? For performance-focused PTO users, we might consider recommending that they pass element indices directly from the frontend and use the `gather_mask` interface instead.

LLMZhangYC (approved):
lgtm

fuhouyu-hw (approved):
/lgtm

PR #1015, opened 2026-05-13
[PTO] Unify TileMatL0A with platform #ifdef instead of separate TileMatL0A_a5

handsomeRobotSK:
Unify TileMatL0A with platform #ifdef instead of separate TileMatL0A_a5

Use #ifdef PTO_PLATFORM_A5 to select BLayout (ColMajor on A5, RowMajor otherwise) within a single TileMatL0A alias, avoiding cascading platform checks in every interface that references it.


LLMZhangYC (approved):
lgtm

benyang0506 (approved):

PR #1067, opened 2026-05-25
[Fix] Update TEXPANDS call to use resolved destination slice name in FillCodegen.

ArmandAlbert:
Update TEXPANDS call to use resolved destination slice name in FillCodegen. Now mhc can pass.


fuhouyu-hw (approved):
/lgtm

LLMZhangYC (approved):
approve

PR #1089, opened 2026-05-28
[Feat] Support KERNEL_TYPE_AIV_ONLY for pure vector kernels in Ascend C backend 🤖

lwwangcgz:
## Summary

Support automatic kernel type selection for Ascend C backend: pure vector kernels use `KERNEL_TYPE_AIV_ONLY` to reduce launch overhead, while kernels with cube operations continue using `KERNEL_TYPE_MIX_AIC_1_1/1_2`.

## Changes

- Add `CubeOpDetector` visitor to detect gemm/mma operations in kernel body
- Generate `KERNEL_TYPE_AIV_ONLY` for pure vector kernels (no cube ops detected)
- Generate `KERNEL_TYPE_MIX_AIC_1_1/1_2` for kernels with cube operations (unchanged behavior)
- Handle vid derivation from block ID (`__raw_cid__ % 2`) for pure vector + cv_1_2 mode
- Multiply grid size by 2 for pure vector + cv_1_2 to replace sub-block mechanism
- Use `SyncAll<true>` for pure vector kernels to avoid cross-core sync overhead

## Background

Previously, the Ascend C backend always generated `KERNEL_TASK_TYPE_DEFAULT(KERNEL_TYPE_MIX_AIC_1_1/1_2)`, causing unnecessary AIC core allocation for pure vector operators. This resulted in higher kernel launch overhead and wasted hardware resources.

The PTO backend already supports automatic kernel type inference by the compiler, so no changes are needed there.

## Testing

Verified with msprof profiling:
- Pure vector kernel (elementwise_add): `main_kernel` now shows `AI_VECTOR_CORE` (was `MIX_AIC`)
- GEMM kernel: `main_kernel` still shows `MIX_AIC` (correct, has cube ops)
- All existing examples pass: elementwise_add, gemm, activation, normalization, reduce, etc.

## Files Changed

- `src/target/codegen_ascend.h`: Add `has_cube_op_` member variable
- `src/target/codegen_ascend.cc`: Add `CubeOpDetector` and kernel type selection logic

### DIFF
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -692,10 +692,6 @@ void CodeGenTileLangAscend::VisitStmt_(const AttrStmtNode *op) {
       }
       this->stream << "auto " << current_block_id
                    << " = AscendC::GetBlockIdx();\n";
-      if (!has_cube_op_ && cv_ratio_ != cv_1_1) {
-        this->PrintIndent();
-        this->stream << "auto __raw_cid__ = " << current_block_id << ";\n";
-      }
       this->PrintIndent();
       this->stream << "if ASCEND_IS_AIV {\n";
       this->PrintIndent();
@@ -714,7 +710,7 @@ void CodeGenTileLangAscend::VisitStmt_(const AttrStmtNode *op) {
       auto vec_id_ = AllocVarID(iv->var.get());
       this->PrintIndent();
       if (!has_cube_op_ && cv_ratio_ != cv_1_1) {
-        this->stream << "auto " << vec_id_ << " = __raw_cid__ % 2;\n";
+        this->stream << "auto " << vec_id_ << " = AscendC::GetBlockIdx() % 2;\n";
       } else {
         this->stream << "auto " << vec_id_ << " = AscendC::GetSubBlockIdx();\n";
       }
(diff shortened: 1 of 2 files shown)


lwwangcgz:
https://github.com/tile-ai/tilelang-ascend/issues/1046

src/target/codegen_ascend.cc
```diff
@@ -672,14 +704,17 @@ void CodeGenTileLangAscend::VisitStmt_(const AttrStmtNode *op) {
...
+                iv->thread_tag == "threadIdx.x") &&
+               iv->var->name_hint != "_") {
       auto vec_id_ = AllocVarID(iv->var.get());
       this->PrintIndent();
-      this->stream << "auto " << vec_id_ << " = AscendC::GetSubBlockIdx();\n";
-    } else if (iv->thread_tag == "threadIdx.x") {
-      auto vec_id_ = AllocVarID(iv->var.get());
-      this->PrintIndent();
-      this->stream << "auto " << vec_id_ << " = AscendC::GetSubBlockIdx();\n";
+      if (!has_cube_op_ && cv_ratio_ != cv_1_1) {
```
benyang0506, line 712: Good job! Could we add some codegen/runtime tests for this change? It touches the core codegen path, including kernel type selection and launch block config.

It would be good to cover cases like:

* Pure AIV kernels generate KERNEL_TYPE_AIV_ONLY
* AIV-only + cv_ratio_ != cv_1_1 doubles the launch block count
* Mixed kernels don’t accidentally go through the AIV-only path

For codegen tests, we can directly check the generated code, e.g. pure AIV cases should contain KERNEL_TYPE_AIV_ONLY.

PR #1091, opened 2026-05-29
[Fix] Fixed the faulty slicing logic of pto codegen.

ArmandAlbert:
Fixed the faulty slicing logic; Now flash_attn_bhsd_auto_pipeline_h16_d128 can pass correctly.

**Issue 1: Error in handling sliced ​​UB buffers for binary vector scalar operations**
When the operands of binary vector scalar operations such as TADDS, TMULS, TSUBS, and TDIVS are sub-regions (slices) of the UB buffer, the original code directly uses the full buffer name when generating the PTO, without resolving the slice semantics, causing the computation to operate in the wrong memory region. Solution: Detect slice operands and generate a TileUbDataND temporary variable with the correct dimensions.

**Issue 2: Incorrect Overriding of `slice_valid_col` in the `clear=True` Path for `reduce`**
In the `clear=True` branch of row/column reduce, the original code incorrectly overridden `dst.slice_valid_col`: during row reduce, `buffer_slice_row` (number of rows) was used to cover the effective column width, and during column reduce, redundant assignments were made. This caused the reduce calculation area to be completely incorrect. Solution: Delete the two incorrect `if` blocks and use the correctly initialized slice information at the beginning of the function.

### DIFF
--- a/src/target/codegen_ascend_pto.cc
+++ b/src/target/codegen_ascend_pto.cc
@@ -2035,12 +2035,26 @@ void CodeGenTileLangAscendPto::BinaryVecOpsCodegen(const CallNode *op,
 
   if (!buffer) {
     std::string scalar = apply_scalar_for_half(index);
-    this->PrintIndent();
-    this->stream << operation << "(";
-    for (const auto &name : var_names) {
-      this->stream << name << ", ";
+
+    ShapeInfo src_info = GetSliceInfo(op->args[1].as<CallNode>());
+    ShapeInfo dst_info = GetSliceInfo(op->args[0].as<CallNode>());
+
+    if (src_info.is_slice || dst_info.is_slice) {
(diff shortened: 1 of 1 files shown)


src/target/codegen_ascend_pto.cc
```diff
@@ -2035,6 +2035,21 @@ void CodeGenTileLangAscendPto::BinaryVecOpsCodegen(const CallNode *op,
 
   if (!buffer) {
     std::string scalar = apply_scalar_for_half(index);
+
```
LLMZhangYC, line 2038: Write the root cause of this issue in the PR description.

ArmandAlbert: Updated.

fuhouyu-hw (approved):
/approve

LLMZhangYC (approved):
approve

PR #1119, opened 2026-06-03
[Fix] Improve scalar handling for complex expressions in BinaryVecOpsCodegen.

ArmandAlbert:
Improve scalar handling for complex expressions in BinaryVecOpsCodegen.

### DIFF
--- a/src/target/codegen_ascend_pto.cc
+++ b/src/target/codegen_ascend_pto.cc
@@ -2091,34 +2091,25 @@ void CodeGenTileLangAscendPto::BinaryVecOpsCodegen(const CallNode *op,
   }
 
   std::string buf_offset = PrintBufferOffset(buffer);
-  std::string scalar_name = GetTempVarName(buf_offset + "_scalar");
+  std::string temp_name = GetTempVarName(buf_offset + "_scalar");
 
   this->PrintIndent();
   this->stream << "set_flag(PIPE_V, PIPE_S, EVENT_ID0);\n";
   this->PrintIndent();
   this->stream << "wait_flag(PIPE_V, PIPE_S, EVENT_ID0);\n";
   this->PrintIndent();
-  this->stream << "auto " << scalar_name << " = " << buf_offset << ".GetValue("
+  this->stream << "auto " << temp_name << " = " << buf_offset << ".GetValue("
                << index << ");\n";
 
-  std::string applied_scalar = apply_scalar_for_half(scalar_name);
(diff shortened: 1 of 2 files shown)


src/target/codegen_ascend_pto.cc
```diff
@@ -2053,7 +2053,19 @@ void CodeGenTileLangAscendPto::BinaryVecOpsCodegen(const CallNode *op,
   auto buffer = op->args[2].as<CallNode>();
 
   if (!buffer) {
-    std::string scalar = apply_scalar_for_half(index);
+    std::string scalar_expr = index;
+    if (IsComplexExpression(op->args[op->args.size() - 2])) {
+      std::string temp_name = GetTempVarName("complex_scalar");
+      this->PrintIndent();
+      this->stream << "set_flag(PIPE_V, PIPE_S, EVENT_ID0);\n";
```
fuhouyu-hw, line 2060: 在codegen插同步是否合理？

ArmandAlbert: Adding synchronization where buffer copying is involved is necessary, and it also exists in the original logic of pto codegen.
The buffer read by temp_name may be being written. If synchronization is not performed, dirty data may be read.
Code:
https://github.com/tile-ai/tilelang-ascend/blob/367d9dd388abb2665be183af0004af46aa393766/src/target/codegen_ascend_pto.cc#L2089-L2090

Example1:
https://github.com/tile-ai/tilelang-ascend/blob/367d9dd388abb2665be183af0004af46aa393766/src/target/codegen_ascend_pto.cc#L1417-L1429

Example2:
https://github.com/tile-ai/tilelang-ascend/blob/367d9dd388abb2665be183af0004af46aa393766/src/target/codegen_ascend_pto.cc#L2081-L2094

fuhouyu-hw (approved):
/lgtm

src/target/codegen_ascend_pto.cc
benyang0506: 这个同步语句可以移到tilelang前端或者pass中吗

ArmandAlbert: The sync for temp_buffer can move to auto_sync pass, but it`s combined with temp_buffer, all logics of temp_buffer need to be done in early pass, so that syncs can be processed by pass rather than codegen.
This requirement involves many changes, needs to be completed uniformly, and requires planning.

benyang0506 (approved):
lgtm

PR #1157, opened 2026-06-11
[Fix] Fix PTO/AscendC Transpose API and add pytest example.

ArmandAlbert:


fuhouyu-hw (approved):
/lgtm

ChaoyangJi:
目前实现为功能实现，性能会比较差，后续需要优化。

ArmandAlbert:
The Ascendc interface will be implemented using a high-performance interface in the future.

ChaoyangJi (approved):

testing/python/language/test_tilelang_ascend_language_elementwise.py
```diff
@@ -4298,6 +4298,14 @@ def test_transpose(dtype, target, shape):
     run_test_transpose(M, N, 16, 16, dtype, target)
 
 
+@pytest.mark.parametrize("dtype", ["float16", "float", "int16", "int32"])
+@pytest.mark.parametrize("target", ["ascendc", "pto"])
+@pytest.mark.parametrize("shape", [(32, 32), (64, 64), (32, 16), (16, 32), (64, 32), (32, 64)])
```
benyang0506, line 4303: What happens if the shape is (24, 40)? If this scenario is unsupported, an assertion should be added instead of returning incorrect results. This case will likely be supported eventually.

lwwangcgz:
#1158

PR #1180, opened 2026-06-15
disable README primary agent

yuqiha:
If disable: true is not added, the readme.md file will be recognized as the primary agent.
image


benyang0506 (approved):
lgtm

fuhouyu-hw (approved):

Issue #1176, opened 2026-06-15
[BUG] worksapce消除与combineCV冲突


fengz72:
## 环境
commit-id: b5f7806fb6a188158c85ea1f3df359494bef93e7
cann: 8.5.0
npu: A3

## kernel
```python
import argparse

import tilelang
import tilelang.language as T
import torch

tilelang.cache.clear_cache()

pass_configs = {
    tilelang.PassConfigKey.TL_ASCEND_AUTO_CV_COMBINE: True,
    tilelang.PassConfigKey.TL_ASCEND_AUTO_SYNC: True,
    tilelang.PassConfigKey.TL_ASCEND_MEMORY_PLANNING: True,
    tilelang.PassConfigKey.TL_ASCEND_AUTO_CV_SYNC: True,
}

parser = argparse.ArgumentParser(description="NPU Kernel Compilation")
parser.add_argument("--m", type=int, default=1024, help="Matrix M dimension")
parser.add_argument("--n", type=int, default=1024, help="Matrix N dimension")
parser.add_argument("--k", type=int, default=1024, help="Matrix K dimension")
args = parser.parse_args()

M = args.m
N = args.n
K = args.k


@tilelang.jit(out_idx=[2], pass_configs=pass_configs)
def matmul_add(M, N, K, block_M, block_N, block_K, dtype="bfloat16", accum_dtype="float"):
    m_num = M // block_M
    n_num = N // block_N

    @T.prim_func
    def main(
        A: T.Tensor((M, K), dtype),
        B: T.Tensor((K, N), dtype),
        C: T.Tensor((M, N), dtype),
        D: T.Tensor((M, N), dtype),
    ):
        with T.Kernel(m_num * n_num, threads=2, is_npu=True) as (cid):
            bx = cid // n_num
            by = cid % n_num
            A_L1 = T.alloc_shared((block_M, block_K), dtype)
            B_L1 = T.alloc_shared((block_K, block_N), dtype)

            C_L0 = T.alloc_fragment((block_M, block_N), accum_dtype)

            d_ub = T.alloc_shared((block_M, block_N), dtype)
            c_ub = T.alloc_shared((block_M, block_N), dtype)

            d_ub_float = T.alloc_shared((block_M, block_N), accum_dtype)
            c_ub_float = T.alloc_shared((block_M, block_N), accum_dtype)

            loop_k = T.ceildiv(K, block_K)
            for k in T.serial(loop_k):
[...]

PR #1181, opened 2026-06-15
[Fix] Fix workspace reduction type_map_ keys and copy-back insertion 🤖

liggest:
## Summary

Fix two bugs in `AscendWorkspaceReduction` pass that caused conflicts with `CombineCV`, resulting in sync point mismatches.

Fixes the half of https://github.com/tile-ai/tilelang-ascend/issues/1176

### 1. Fix `type_map_` keys to match actual Ascend C type names

The `CopyInfoCollector::type_map_` used incorrect key names (e.g. `int8`, `uint32`) that didn't match the actual C++ type strings produced by codegen (e.g. `int8_t`, `uint32_t`). Also added missing entries for `bfloat16_t` and `AscendC::int4b_t`.

### 2. Treat copy calls as possible data consumers to correctly insert copy-backs

The previous logic blanket-skipped all copy-related call statements, assuming they were always producers. This caused the pass to miss cases where a copy call *consumes* a workspace buffer (read access), leading to missing copy-back insertions and downstream sync mismatches in `CombineCV`.

Replaced the coarse copy-statement skip with fine-grained `rw_mask` checking on `access_ptr` arguments: only write-only accesses (`rw_mask == 2`) are treated as producers; read accesses correctly trigger copy-back insertion.

## Testing

Verified with the reproducer kernel from #1176 (matmul_add with `threads=2`, `bfloat16`, all auto-sync/memory-planning passes enabled) — the mentioned issue should be solved.

### DIFF
--- a/src/transform/ascend_workspace_reduction.cc
+++ b/src/transform/ascend_workspace_reduction.cc
@@ -616,9 +616,10 @@ private:
 
       // access_args layout: [type_annotation, var, offset, extent, rw_mask]
       // Only consider buffer as a "consumer" if rw_mask has read bit set.
-      // Skip write-only access (rw_mask == 2) since no copy-back needed before a write.
-      ICHECK(access_args.size() >= 5)
-          << "[Error]<WorkspaceReduction>: access ptr args size too small for rw_mask!";
+      // Skip write-only access (rw_mask == 2) since no copy-back needed before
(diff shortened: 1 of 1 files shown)


z00520135 (approved):

benyang0506 (approved):
lgtm

PR #1189, opened 2026-06-16
[feat]Add a universal script for precision testing

LLMZhangYC:


benyang0506:
Could you clarify the source of this accuracy standard in the PR description? This would help the community better understand the rationale behind the standard.

Issue #1194, opened 2026-06-16
[Bug]  Incorrect `T.Parallel` lowering for compact-to-aligned UB store


Hibbert133:
## Bug Description

When using `T.Parallel` to copy data from a compact UB layout to an aligned/padded UB layout, the Ascend backend may incorrectly lower the loop into a full contiguous `copy_ub_to_ub`.

The source and destination have different row strides, so this rewrite is not semantically equivalent. In my case, this leads to wrong-layout UB reads and incorrect numerical results.

There is also a related issue: a 1D `T.Parallel` workaround with several scalar UB stores can fail during codegen with `Find undefined Variable v_thread`.

## Environment

  Platform: Ascend 910B2
  Pass config:
    TL_ASCEND_AUTO_SYNC = False
    TL_ASCEND_MEMORY_PLANNING = True

## Problematic pattern

The simplified pattern is:

```python
# src_compact: [G, 4]
# dst_aligned: [G, 8]
# only the first 4 entries of each row are valid
# the last 4 entries are padding for aligned reduce

T.tile.fill(dst_aligned, 0)

for (group, inner) in T.Parallel(G, 4):
    dst_aligned[group, inner] = src_compact[group, inner]
```

Expected semantics:

```text
dst_aligned[group, 0:4] = src_compact[group, 0:4]
dst_aligned[group, 4:8] remains 0
```

This pattern is useful when a later operator, such as `reduce_max`, requires or performs better with an aligned inner dimension.

---

## Actual generated C++

The Ascend backend lowers the above loop into a full contiguous copy like:

```cpp
tl::ascend::Fill<int>(dst_aligned[0], 0, 2048);
tl::ascend::copy_ub_to_ub<int, int, 2048>(dst_aligned[0], src_compact[0]);
```

This is incorrect because `src_compact` only contains the compact `[G, 4]` data, while `dst_aligned` has the padded `[G, 8]` layout.

So the generated code effectively assumes:

```text
src_compact has the same flattened size and row stride as dst_aligned
```

but that is not true.

---

## Why this is wrong

For the original loop, the intended address mapping is:

```text
dst_index = group * 8 + inner
src_index = group * 4 + inner
```

for `inner in [0, 4)`.

The generated contiguous copy behaves like:

```text
dst_index = flat
src_index = flat
```

for the full aligned size.

These two mappings are only equivalent if the source and destination row strides are the same. Here they are different, so the rewrite corrupts the layout.

---

## Observed behavior

[...]

PR #1215, opened 2026-06-22
[fix] fix T.Parallel UB copy lowering

mygitljf:
This PR fixes issue #1194, where `T.Parallel` could lower a compact-to-aligned UB store into an incorrect flat UB-to-UB copy.

The problematic case is copying compact logical data like `[G, 4]` into the first columns of an aligned UB buffer like `[G, 8]`. That copy must preserve the destination row stride, otherwise data from later rows can be written into padding or adjacent groups.

Changes:
- Preserve compact logical copy extents for source-indexed 2D UB copies.
- Preserve the destination offset instead of resetting destination indices to zero.
- Emit PTO `copy_ub_to_ub_strided` when physical row stride differs from logical copy width.
- Derive physical UB row stride from slice metadata in PTO codegen.
- Add shape checks for UB-to-UB source/destination tile compatibility.
- Replace fallback `v_thread` with constant `0` when no real thread axis exists.
- Add regression tests for compact-to-aligned UB copy and `v_thread` codegen.

## Validation
Local validation:
- `ruff check` passed for the new regression test.
- Python compile check passed for the new regression test.
- `git diff --check` passed.
- C++ syntax checks passed for the three changed C++ files.

Remote Ascend validation:
- Rebuilt TileLang on the Ascend server.
- Verified `tilelang` and `tvm` import from the rebuilt tree.
- Ran targeted issue #1194 pytest.
- Result: `2 passed, 2 skipped`.

Notes:
- The two passing tests verify source/codegen behavior:
  - compact-to-aligned UB copy emits strided PTO copy
  - generated source no longer references `v_thread`
- The two skipped tests are runtime/NPU behavior checks.
- They were skipped because the remote PTO JIT compiler cannot compile a minimal `pto-isa` probe in the current environment.
- The skip is environment-related; skipped runtime tests are not counted as passed.

### DIFF
--- a/src/target/codegen_ascend_pto.cc
+++ b/src/target/codegen_ascend_pto.cc
@@ -1172,13 +1172,20 @@ void CodeGenTileLangAscendPto::CopyUBToUBCodegen(const CallNode *call) {
     auto dst_tile_cols = Downcast<IntImm>(call->args[7])->value;
     auto dst_buf_cols = Downcast<IntImm>(call->args[8])->value;
 
+    ICHECK_EQ(src_tile_rows, dst_tile_rows)
+        << "UB-to-UB copy requires matching source and destination row counts";
+    ICHECK_EQ(src_tile_cols, dst_tile_cols)
+        << "UB-to-UB copy requires matching source and destination column counts";
+
+    ShapeInfo src_shape_info = GetSliceInfo(src_info.access_ptr);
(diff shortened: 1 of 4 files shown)


benyang0506:
Thanks for your contribution! The issue has been fixed in https://github.com/tile-ai/tilelang-ascend/pull/1219. Please pull the latest code and give it a try.

PR #1219, opened 2026-06-23
[Fix] Lower un-vectorizable T.Parallel to serial (fixes #1194)

### LINKED ISSUE #1194: [Bug]  Incorrect `T.Parallel` lowering for compact-to-aligned UB store

fuhouyu-hw:
## Summary

Fixes #1194. A `T.Parallel` element-wise loop that cannot be safely vectorized
produced wrong results or failed to compile, instead of falling back to a
correct scalar loop. Two symptoms, one root cause.

**Compact → aligned UB store** (different row/inner width), e.g.

```python
T.tile.fill(dst, 0)                  # dst: [G, 16]
for g, inner in T.Parallel(G, 8):    # src: [G, 8]
    dst[g, inner] = src[g, inner]
```

was lowered to a `copy_ub_to_ub` whose source width was taken from the
destination, packing two source rows into one destination row and producing
wrong results.

**Un-vectorizable body** (e.g. a data-dependent `if/else`) left the original
`kParallel` loop unchanged, which then tripped
`InternalError: Find undefined Variable v_thread` at codegen.

## Fix

Ascend has no native parallel-for, so any `T.Parallel` that cannot be safely
vectorized is now lowered to a plain serial loop (codegen emits element-wise
`SetValue` / `GetValue`), which is correct for arbitrary row strides and free
of the `v_thread` issue:

- Refuse to vectorize a UB→UB copy whose inner (row) dimensions differ.
- Propagate that bail for UB outputs, not only GM outputs.
- At both `kParallel` fall-through points (1D and 2D), lower the loop(s) to
  `kSerial` instead of leaving a `kParallel` for codegen.

`copy_ub_to_ub` emission, the strided template, and `ascend.cc` are left
untouched, so existing GM↔UB / matmul copy paths are unaffected.

## Testing

`testing/python/language/parallel/test_tilelang_ascend_language_parallel_ub_stride.py` (Ascend NPU):

- compact → aligned with 32B-aligned row widths (int32 8→16 / 16→32, float16
  16→32);
- same-width control (16→16);
- 1D data-dependent if/else falling back to a scalar relu.

All pass on Ascend 910B2. The full `testing/python/language/parallel/`
regression suite is green.

Closes #1194

### DIFF
--- a/src/transform/ascend_lower_parallel_to_vector.cc
+++ b/src/transform/ascend_lower_parallel_to_vector.cc
@@ -284,7 +284,14 @@ private:
       // parallel -> parallel -> (store | seq)
       const auto *inner_for = op->body.as<ForNode>();
       if (!inner_for || inner_for->kind != ForKind::kParallel) {
-        return arith::IRMutatorWithAnalyzer::VisitStmt_(op);
+        // Not a 2D parallel nest and the 1D path above could not vectorize it:
+        // lower this parallel loop to serial. Ascend has no native
+        // parallel-for, so a surviving kParallel would reach codegen and trip
+        // "Find undefined Variable v_thread".
+        auto serial = make_object<ForNode>(*op);
+        serial->kind = ForKind::kSerial;
+        serial->body = VisitStmt(op->body);
+        return Stmt(serial);
       }
 
       const auto *third_for = inner_for->body.as<ForNode>();
@@ -316,7 +323,16 @@ private:
           return v;
       }
 
-      return arith::IRMutatorWithAnalyzer::VisitStmt_(op);
+      // 2D parallel nest that could not be vectorized (e.g. a compact->aligned
+      // UB copy): lower both levels to serial instead of leaving kParallel
+      // loops for codegen.
+      auto outer = make_object<ForNode>(*op);
+      outer->kind = ForKind::kSerial;
+      auto inner = make_object<ForNode>(*inner_for);
+      inner->kind = ForKind::kSerial;
+      inner->body = VisitStmt(inner_for->body);
+      outer->body = Stmt(inner);
+      return Stmt(outer);
     }
 
     // Serial cases
@@ -756,7 +772,12 @@ private:
         parallel_vars, &row_stmts, is_2d, inner_vec_len);
 
     is_2d_vectorizing_ = saved_is_2d_vectorizing;
-    if (is_global_output && (!success || row_stmts.empty()))
+    // A failed decomposition (e.g. a compact->aligned UB->UB copy that cannot
+    // be lowered to copy_ub_to_ub) must bail regardless of the output scope so
+    // VisitStmt_ falls back to a scalar serial loop.
+    if (!success)
+      return NullOpt;
+    if (is_global_output && row_stmts.empty())
       return NullOpt;
 
     // If output is in GM, add ascend_copy to copy from temp UB to GM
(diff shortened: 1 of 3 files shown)


LLMZhangYC (approved):
lgtm

benyang0506 (approved):
lgtm

PR #1223, opened 2026-06-24
[Test] Add tail-block (尾块) guard suite for CUBE/VECTOR/CV operators

fuhouyu-hw:
Guard tests for the framework's automatic tail-block handling on non-divisible shapes, which the existing divisible-shape suites never exercise. Covers, on both the ascendc and pto backends:

  * CUBE  (gemm)           : gm2l1 + l0c2gm clamp, M/N/K tails
  * VECTOR (add / abs)     : gm2ub + ub2gm clamp, M/N tails
  * VECTOR (reduce_max)    : sliced reduce via real_shape
  * CV fusion (matmul+add) : C-scope cube + V-scope vector tails

All configs use deliberately non-divisible M/N/K so the frontend-transparent tail path (compute_valid_extent in src/op/ascend.cc) is exercised. Requires NPU hardware to execute.

### DIFF
--- a/testing/python/language/test_tilelang_ascend_language_tail_block.py
+++ b/testing/python/language/test_tilelang_ascend_language_tail_block.py
@@ -1,0 +1,424 @@
+import pytest
+import tilelang
+import tilelang.language as T
+import torch
+
+"""
+Tail-block (尾块) guard suite.
+
+Feature under test
+------------------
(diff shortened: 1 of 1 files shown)


benyang0506 (approved):
lgtm

LLMZhangYC (approved):
approve

PR #1243, opened 2026-06-26
[Fix] Fix Ascendc bug for n-way GM->L1 Nz layout slice copy.

ArmandAlbert:
**Fix Ascendc slice copy bug.**
Only the primary copy (dst offset 0, i.e. `need_clear == true`) is allowed to zero-init the full L1 tile. Sub-region copies (`need_clear == false`, e.g. the second DMA of a splice / vertical-merge pattern) must NOT clear, otherwise they clobber data already written into the same NZ tile. The full-tile clear is only correct when it targets the tile base, which the codegen guarantees by passing `need_clear = (dst_offset == 0)`.

## Completion Summary
I have fully implemented the feature of **"multi-block discontinuous GM → single L1 unified NZ layout"** in the AscendC backend branch. Both `full_copy_gemm` and `splice_copy_gemm` have passed verification.

## Root Cause
The helper function `copy_gm_to_l1` uses `InitConstValue` to zero out the entire `dstM*dstN` L1 block when `realTailM < dstM` (partial copy). For the second DMA in the splice flow (`k_l1[1024]`), this zeroing operation writes 16384 elements starting at offset 1024, overwriting the data already written by the first DMA, resulting in corrupted NZ layout (max_diff=50).

Note that codegen has already generated the correct NZ offset (`1024 = 64*C0_SIZE`) and the overall NZ framework (template `<128, 128>` → `dstNzC0Stride=128`). The only defect is this unconditional zeroing.

### `src/tl_templates/ascend/common.h`
Added a default parameter `bool need_clear = true` to the helper, and modified the zeroing condition to `need_clear && (tailM != dstM || tailN != dstN)`. The default value `true` preserves the original behavior for all direct C++ callers.

### `src/target/codegen_ascend.cc`
In CopyCodegen, the synthesized boolean `need_clear` is appended after the 3 standard extra arguments for `copy_gm_to_l1`:
`is_zero(dst_offset_expr) ? "true" : "false"`

Specifically, zeroing the entire block is only permitted when the DMA target is located at the tile start (offset 0); zeroing is skipped for sub-region copies with non-zero offsets. No changes to the op's argument structure are required (`args[6]` and `args[7]` are already dead parameters), and the `workspace_reduction` path also requires no modification — its copy is full, so `realTailM=0→dstM` will never trigger zeroing.

### DIFF
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -2344,6 +2344,17 @@ void CodeGenTileLangAscend::CopyCodegen(const CallNode *op) {
       this->stream << ", " << var_names[i];
     }
 
+    // copy_gm_to_l1: append a `need_clear` flag so the helper only zero-inits
+    // the full L1 tile for the primary copy (dst offset 0). Sub-region copies
+    // (non-zero dst offset, e.g. the second DMA of a splice / vertical-merge
+    // pattern) must skip the clear, otherwise they clobber data already written
+    // into the same NZ tile. dst_offset_expr is the scalar start offset of the
+    // destination tile; when it is 0 this DMA targets the tile base and is the
+    // natural owner of the tail-padding clear.
+    if (op_name.find("copy_gm_to_l1") != std::string::npos) {
+      this->stream << ", " << (is_zero(dst_offset_expr) ? "true" : "false");
+    }
+
     this->stream << ");\n";
   } else {
     this->PrintIndent();
--- a/src/tl_templates/ascend/common.h
+++ b/src/tl_templates/ascend/common.h
@@ -56,10 +56,17 @@ template <typename T, uint32_t dstM, uint32_t dstN>
 CATLASS_DEVICE void copy_gm_to_l1(LocalTensor<T> dstTensor,
                                   GlobalTensor<T> srcTensor,
                                   uint32_t realSrcN = 1, uint32_t realTailM = 0,
-                                  uint32_t realTailN = 0) {
+                                  uint32_t realTailN = 0,
+                                  bool need_clear = true) {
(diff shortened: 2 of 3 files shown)


ArmandAlbert:
Created test file testing/python/language/test_tilelang_ascend_language_gm_to_l1_splice.py with a total of 49 test cases, all passed, covering 9 groups of GM→L1 scenarios:
<html>
<body>
<html><head>

Group | Scenario | Operator | Layout / Transpose | Variable Type | Number of Cases
-- | -- | -- | -- | -- | --
G1 | Full Block Copy Baseline | full_copy_annotated / full_copy_plain | zN+nZ / default zN; transB=T/F | Constant | 4
G2 | K-tail Zeroing Regression | full_copy_plain | Default zN; transB=F | Constant | 3
G3 | 2-way nZ splice | splice_2way_nz | nZ; transB=T | Constant | 12
G4 | 3-way splice + Unfilled Tail | splice_3way_tail | nZ; transB=T | Constant | 4
G5 | 2-way zN splice | splice_2way_zn | zN; transB=F | Constant | 3
G6 | Combined splice + K-tail | splice_with_k_tail | nZ; transB=T | Constant | 2
G7 | Dynamic Batch + Dynamic Split | splice_2way_nz_dynamic_batch | nZ; transB=T | Symbolic batch + split | 12
G8 | Dynamic K K-tail | full_copy_plain_dynamic_k | Default zN; transB=F | Symbolic K | 3
G9 | Dynamic split splice | splice_2way_nz_dynamic_split | nZ; transB=T | Symbolic split | 6

LLMZhangYC (dismissed):
lgtm

fuhouyu-hw (approved):
lgtm

ChaoyangJi (approved):
approve

PR #1258, opened 2026-06-29
[Fix] Stop tracking generated AOT example artifacts 🤖

platelett:
## Summary

Generated AOT example artifacts dirty `git status` on every run. This PR removes
one at its source and ignores the other:

- The `ascendc.cpp` kernel-source dump was written purely for inspection/debugging,
  so it is now **removed at its source** rather than ignored (per review feedback).
- `examples/gemm_aot/kernel_lib.so` is a genuine build output, so it stays **ignored**.

This PR owns **all** the generated-artifact ignore rules (see *Coordination with
#1276* below) so only one PR edits `.gitignore` — the two no longer collide on the
same file-tail anchor and can merge in any order without conflict.

## Changes

- **`examples/fused_sigmoid_gating_delta_rule/fused_sigmoid_gating_delta_rule_varlen.py`**
  — drop the debug-only file I/O:

  ```python
  with open("./ascendc.cpp", "w") as fp:
      fp.write(ker.get_kernel_source())
  ```

  The dump was inspection-only and never read back, so removing it stops the
  artifact from being produced — no ignore rule needed for it.

- **`.gitignore`** (mirrors the existing `docs/notebook/*.so` rule) — ignore
  `examples/gemm_aot/kernel_lib.so`, compiled by the reworked gemm_aot AOT flow in #1276.

> `examples/gemm_aot/example_gemm.cpp` is no longer generated — the new flow in
> #1276 compiles straight to `kernel_lib.so` and untracks the orphaned `.cpp`
> there — so it needs no ignore rule here.

## Coordination with #1276

#1276 originally also added the `kernel_lib.so` ignore rule, which collided with
this PR at the same `.gitignore` tail anchor. To let the two merge in any order
without conflict, #1276 drops its `.gitignore` change and this PR owns the rule.

## Testing

The deleted lines only wrote `ker.get_kernel_source()` to `./ascendc.cpp` **after**
the kernel was already built, and nothing reads that file back — so kernel
construction, execution and the golden comparison are all unaffected. The example's
own accuracy gate still validates the change end to end:

```bash
python examples/fused_sigmoid_gating_delta_rule/fused_sigmoid_gating_delta_rule_varlen.py
# per case: assert_close(out / final_state, golden)  ->  "Kernel Output Match!"
# all cases:                                             "All tests passed!"
# and ./ascendc.cpp is no longer written into the working tree
```


LLMZhangYC (dismissed):
lgtm

.gitignore
```diff
@@ -104,3 +104,9 @@ docs/notebook/*.cpp
 
 .agents/reports
 
+# gemm_aot AOT build output
+examples/gemm_aot/kernel_lib.so
+
+# Kernel-source debug dump written via get_kernel_source()
+examples/fused_sigmoid_gating_delta_rule/ascendc.cpp
```
benyang0506: It seems that:
```
with open("./ascendc.cpp", "w") as fp:
    fp.write(ker.get_kernel_source())
```
is purely for checking and debugging. I think this file I/O should be removed in:
https://github.com/tile-ai/tilelang-ascend/blob/ascendc_pto/examples/fused_sigmoid_gating_delta_rule/fused_sigmoid_gating_delta_rule_varlen.py#L425

liggest:
Looks good!💪
These untracked commit-discouraged files always feel troublesome after running a bench test locally.

benyang0506 (approved):
lgtm

LLMZhangYC (approved):
approve

PR #1271, opened 2026-06-30
[Ascend] Vector tail-block scheme — batch 1 (remove front-end pad_value, valid-region compute)

fuhouyu-hw:
## What
Replace the front-end `pad_value` tail handling on the AscendC backend with a
backend-internal "propagate valid shape + compute only the valid region"
mechanism, so `T.copy` matches upstream tilelang semantics.

`copy_gm_to_ub` records the valid `[row,col]` rect on the dst UB buffer → new
pass `AscendTailMaskPropagation` (after `LowerTileOp`) propagates it → vector ops
on a tail buffer are rewritten to `tl.ascend_tail_*` with runtime
valid_row/valid_col/physical_col → codegen emits `tl::ascend::tail_*` helpers
that compute only the valid rectangle.

## Scope (batch 1)
- Rewritten: unary / binary / scalar / reduce.
- Propagate-only (per-lane, still correct full-tile): cast / broadcast / ub→ub.
- Provably-full copies stay untouched (IRMutatorWithAnalyzer).
- Deferred to batch 2: compare/select (packed mask), broadcast rewrite.

## Validation
Draft — opening for CI signal (dev env has no Ascend toolchain/NPU).
- Host build ✅
- `examples/tail_mask/example_tail_add.py` (float/float16/100×200): ✅ on NPU
- reduce reworked to `Pattern::Reduce::AR` (re-verifying)
- `examples/softmax/example_online_softmax.py` (M=34/N=130 double tail): pending NPU

### DIFF
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -575,6 +575,14 @@ void CodeGenTileLangAscend::VisitExpr_(const CallNode *op, std::ostream &os) {
     PrintOpCall(op, "AscendC::Xor", {0, op->args.size() - 1}, {0, 0});
   } else if (op->op.same_as(tl::ascend_broadcast())) {
     BroadcastOpCodegen(op);
+  } else if (op->op.same_as(tl::ascend_tail_unary())) {
+    TailUnaryOpCodegen(op);
+  } else if (op->op.same_as(tl::ascend_tail_binary())) {
+    TailBinaryOpCodegen(op);
+  } else if (op->op.same_as(tl::ascend_tail_scalar())) {
+    TailScalarOpCodegen(op);
+  } else if (op->op.same_as(tl::ascend_tail_reduce())) {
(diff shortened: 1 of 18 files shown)


src/target/codegen_ascend.cc
```diff
@@ -575,6 +575,14 @@ void CodeGenTileLangAscend::VisitExpr_(const CallNode *op, std::ostream &os) {
     PrintOpCall(op, "AscendC::Xor", {0, op->args.size() - 1}, {0, 0});
   } else if (op->op.same_as(tl::ascend_broadcast())) {
     BroadcastOpCodegen(op);
+  } else if (op->op.same_as(tl::ascend_tail_unary())) {
```
z00520135, line 578: 新增的这些操作 在ascend_vid_reduction.cc vid消除pass的tile_ops map中 会对基础操作长度做除以2处理，考虑下这些新增操作是否需要加入

PR #1257, opened 2026-06-29
[Fix] Fix gemm_aot AOT example build on CANN 9.1.0 / 9.1.0-beta.1 🤖

platelett:
## Summary

`examples/gemm_aot` builds on older CANN releases but **fails to compile on CANN 9.1.0 and 9.1.0-beta.1**: the profiling headers the Ascend runtime pulls in transitively are not on `build.sh`'s include path. This adds the include roots each version needs.

## Root Cause

`runtime/rt_ffts.h` → `runtime/base.h` does `#include "profiling/prof_api.h"`, which in turn includes `prof_common.h`. Newer CANN relocated these headers to directories the current `build.sh` doesn't search:

- **9.1.0-beta.1** — `profiling/prof_api.h` is under `${ASCEND_HOME_PATH}/pkg_inc/` → `fatal error: 'profiling/prof_api.h' file not found`
- **9.1.0** — `prof_common.h` is under `${ASCEND_HOME_PATH}/include/experiment/msprof/toolchain/` → `fatal error: 'prof_common.h' file not found`

## Changes

`examples/gemm_aot/build.sh` — add both include roots so a single script covers both releases (extra non-matching `-I` paths are harmless):

```
-I${ASCEND_HOME_PATH}/include/experiment/msprof/toolchain   # 9.1.0
-I${ASCEND_HOME_PATH}/pkg_inc                               # 9.1.0-beta.1
```

## Testing

```bash
source set_env.sh
cd examples/gemm_aot
bash run_example_gemm_aot.sh   # builds kernel_lib.so, prints "Kernel Output Match!"
```

Verified on CANN 9.1.0 and 9.1.0-beta.1.

PR #1276, opened 2026-07-01
[Fix] Build gemm_aot AOT example via framework LibraryGenerator 🤖

platelett:
## Summary

`examples/gemm_aot` compiles its kernel with a hand-written `build.sh` that hard-codes the full `bisheng` include/flag set. That list is version-specific and can't be kept correct across CANN releases by hand. This replaces it with the framework's `LibraryGenerator` — the same compile path the JIT flow and the passing example suite already use — so the AOT build resolves the correct include roots per toolkit version automatically.

Supersedes #1257 (which I'll close); background below.

## Background — why #1257 didn't work

- `gemm_aot` builds on older CANN (including CI's **CANN 9.0.0**) but **fails to compile on CANN 9.1.0 and 9.1.0-beta.1**: `runtime/rt_ffts.h → runtime/base.h → profiling/prof_api.h → prof_common.h`, and newer CANN relocated those profiling headers out of `build.sh`'s search path (`prof_api.h` under `pkg_inc/` on 9.1.0-beta.1; `prof_common.h` under `include/experiment/msprof/toolchain/` on 9.1.0).
- #1257 added those two `-I` roots to `build.sh` to fix 9.1.0 / 9.1.0-beta.1, assuming extra non-matching `-I` paths are harmless.
- They weren't: #1257's only change is those two `-I` roots, and on CI's **CANN 9.0.0** the build then stops producing `kernel_lib.so`. The job surfaces only the downstream `OSError: ./kernel_lib.so: cannot open shared object file` — because `run_example_gemm_aot.sh` has no `set -e` and runs the test even after `build.sh` fails. `@z00520135` confirmed CI runs CANN 9.0.0 and the change needs version compatibility.
- Net: no single static `-I` list satisfies 9.0.0, 9.1.0 and 9.1.0-beta.1 at once — fixing one release regresses another. That's the whack-a-mole a hand-maintained `build.sh` invites, and it's what `@liggest` pointed at by suggesting the options be aligned with `libgen.py`.

## Fix

Compile through the framework instead of a bespoke script. `LibraryGenerator` (`tilelang/jit/adapter/libgen.py`) already maintains the correct include roots per toolkit version — it's what every JIT kernel and the passing CI suite compile with.

- **`example_gemm.py`** — after `tilelang.engine.lower(...)`, compile with `LibraryGenerator` and emit `./kernel_lib.so`. Adds `--target {ascendc,pto}` (default `ascendc`), `--platform` (auto-detected), `-o/--output`.
- **`build.sh`** — removed.
- **`run_example_gemm_aot.sh`** — reduced to generate + test, with `set -e` so a compile failure fails loudly instead of masquerading as the test's `OSError`.
- **`README.md`** — updated to the new flow.

`test_example_gemm.py` is unchanged — still loads `./kernel_lib.so` via `ctypes` and calls `lib.call(...)`, preserving the "compile ahead-of-time, run without the JIT stack" demonstration.

> This PR removes the now-orphaned tracked `example_gemm.cpp` (the new flow compiles straight to `kernel_lib.so` and never regenerates it). The `.gitignore` rules for the generated artifacts — including `kernel_lib.so` — are owned by #1258, so this PR no longer touches `.gitignore` and the two won't collide on the same anchor.

## Testing

```bash
source set_env.sh
cd examples/gemm_aot

bash run_example_gemm_aot.sh                                        # default (ascendc): builds ./kernel_lib.so -> Kernel Output Match!
python example_gemm.py --target pto && python test_example_gemm.py # pto backend                              -> Kernel Output Match!
```

Verified on **(A2, CANN 9.1.0)** and **(A3, CANN 9.1.0-beta.1)**; both `ascendc` and `pto` backends print `Kernel Output Match!`.

---

## Appendix — reading `example_gemm.py` without the CRLF → LF noise

`example_gemm.py` shows as ~156 changed lines only because the tracked file was CRLF and the rewrite is LF, so **every** line reads as modified. To review the real content change, open [**Files changed with whitespace hidden**](https://github.com/tile-ai/tilelang-ascend/pull/1276/files?w=1) — `?w=1` treats the CRLF→LF flip as whitespace, so GitHub renders only the actual change, fully styled.


LLMZhangYC (dismissed):
lgtm

benyang0506 (approved):
lgtm

LLMZhangYC (approved):
approve

PR #1255, opened 2026-06-29
[Feat] Xattention and add brcb_experiment , row_expand_{mul,sub,div}_experiment API

liuwenda4:
The performance of xattention tilelang code compared to Ascendc reached 1.16 at beam=128, 0.96 at beam=256, 0.8 at beam=512, and overall 0.8 at Ascendc

Add 4 T. tile experiment APIs and unify AscendC/PTO dual backend calls:
-BRCBExperioment: AscendC emits BRCB, PTO emits TROWEXPAND
-Row_dexpand_stul-experioment: AscendC launches brcb+mul_mask, PTO launches TROWEXPANDMUL_row-vec
-Row_dexpand_Sub_deperiment: AscendC emits brcb+sub_mask, PTO emits TROWEXPANDSUB_row-vec
-Row_dexpand-div-period: AscendC launches brcb+div_mask, PTO launches TROWEXPANDDIV_row-vec

PR #1290, opened 2026-07-03
[Fix] fix MUL/SUB/DIV _experiment    vaildcol != col

liuwenda4:
- row_vec helpers (MUL/SUB/DIV x with/without tmp): decouple dst/src0 physical dims (dstRows/dstCols) from valid dims (dstRowValid/dstColValid), and src1 physical col (Src1Col, deduced) from logical vecLen. Fixes template deduction failure when is_slice buffers have col != validCol.

### DIFF
--- a/src/target/codegen_ascend_pto.cc
+++ b/src/target/codegen_ascend_pto.cc
@@ -2332,14 +2332,17 @@ void CodeGenTileLangAscendPto::RowExpandBinOpExperimentCodegenPto(
     CreateUbVariableND(src0_name, src0);
   }
 
-  int32_t src1_len = src1.slice_col;
   int32_t dst_rows = dst.is_slice ? dst.slice_valid_row : dst.row;
   int32_t dst_cols = dst.is_slice ? dst.slice_valid_col : dst.col;
+  int32_t src1_len = dst_rows;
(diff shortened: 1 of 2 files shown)


liggest:
Background:
After #1255 was merged, its function templates defined in PTO's `common.h` turned up to be incompatible with PTO codegen modification in #1238, causing #1238 fail to pass CI after rebasing. This PR provides a small patch to attempt to fix the issue.

src/target/codegen_ascend_pto.cc
```diff
@@ -3100,15 +3103,17 @@ void CodeGenTileLangAscendPto::VisitStmt_(const AllocateNode *op) {
   ICHECK(shape.size() == 4)
       << "Expected a 4D shape [M, N, Valid_M, Valid_N] for PTO, but got "
       << shape.size() << "D for " << op->buffer_var->name_hint;
-  const auto &M = shape[0];
-  const auto &N = shape[1];
-  const auto &valid_M = shape[2];
-  const auto &valid_N = shape[3];
+  int32_t M_int = shape[0].as<IntImmNode>()->value;
```
liggest: In my opinion, maybe modifying this function is unnecessary.

PR #1218, opened 2026-06-23
[Fix] Fix T.alloc_L1 translated to ub bug.

ArmandAlbert:
Fix T.alloc_L1 translated to ub bug.
Implement L1 and UB explicit splice functions with corresponding tests.
Fix pto codegen GetSliceInfo nullptr break bug.
Fix pto codegen LetStmt param list bug.

Issue #1207, opened 2026-06-18
[Feature] T.tile.fill不支持动态切片


fengz72:
## Feature Description
需要根据运行时计算出来的长度, 动态的指定fill填充的长度
```python
idx = T.if_then_else(A[0, 0] > 0, 32, 64)
T.tile.fill(c_ub[0, 0: idx], 1.0)
```

## Use Case
```
import argparse

import tilelang
import tilelang.language as T
import torch

tilelang.cache.clear_cache()

parser = argparse.ArgumentParser(description="NPU Kernel Compilation")
parser.add_argument("--m", type=int, default=1024, help="Matrix M dimension")
parser.add_argument("--n", type=int, default=1024, help="Matrix N dimension")
args = parser.parse_args()

M = args.m
N = args.n


@tilelang.jit(out_idx=[-1])
def vec_add(M, N, block_M, block_N, dtype="float"):
    m_num = M // block_M
    n_num = N // block_N

    VEC_NUM = 2

    @T.prim_func
    def main(
            A: T.Tensor((M, N), dtype),
            B: T.Tensor((M, N), dtype),
            C: T.Tensor((M, N), dtype),
    ):
        with T.Kernel(m_num * n_num, is_npu=True) as (cid, vid):
            bx = cid // n_num
            by = cid % n_num

            a_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
            b_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
            c_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
            with T.Scope("V"):
                T.copy(A[bx * block_M + vid * block_M // VEC_NUM, by * block_N], a_ub)
                T.copy(B[bx * block_M + vid * block_M // VEC_NUM, by * block_N], b_ub)

                T.barrier_all()
                idx = T.if_then_else(A[0, 0] > 0, 32, 64)
                T.tile.fill(c_ub[0, 0: idx], 1.0)
                T.barrier_all()

                T.copy(c_ub, C[bx * block_M + vid * block_M // VEC_NUM, by * block_N])

    return main


func = vec_add(M, N, 128, 256)

torch.manual_seed(0)

a = torch.randn(M, N).npu()
b = torch.randn(M, N).npu()

torch.npu.synchronize()
print("init successful!")

c = func(a, b)
```

## Proposed Solution
有两个地方需要更改
1. `tilelang/language/ascend_tile.py`
[...]

PR #1295, opened 2026-07-04
[BugFix] Support runtime-dynamic slice length in T.tile.fill (#1207)

fuhouyu-hw:
T.tile.fill(buf[0, 0:idx], v) with a runtime-dynamic idx crashed at parse time with "int() argument must be ... not Var".

Root cause is in the pinned tvm submodule, not tilelang: the TVMScript evaluator auto-fills step=1 for sliced call arguments (buf[0, 0:idx] -> buf[0, 0:idx:1]), and Buffer.__getitem__ treats any non-None step as has_step, routing the slice to the Ramp/BufferLoad path whose int(lanes) requires a compile-time-constant length.

Fix: a 1-line submodule patch so has_step ignores an explicit unit step (a:b:1), keeping such slices on the BufferRegion path. Upstream tvm fixed the same thing in a large refactor that cannot be cherry-picked into the pinned commit, so it is carried as a local patch.

- 3rdparty/patches/tvm_slice_step_fix.patch: the Buffer.__getitem__ fix, plus a README documenting the directory and how patches are applied.
- install_ascend.sh: auto-apply 3rdparty/patches/tvm_*.patch right after submodule init; idempotent, and FATAL on failure so we never build a silently unpatched TVM.
- examples/elementwise/example_dynamic_fill.py: dynamic prefix fill on the default (ascendc) target.
- examples/elementwise/example_dynamic_fill_pto.py: same on target="pto", zero-initialised + exact-length assertion to catch any over-fill.

No tilelang Python change is needed: math.prod already builds a symbolic product for a PrimExpr extent (IntImm defines __rmul__), and the PTO codegen already tolerates a dynamic access extent since #1218.

### DIFF
--- a/install_ascend.sh
+++ b/install_ascend.sh
@@ -173,6 +173,21 @@ echo "Cloning TVM repository and initializing submodules..."
 # clone and build tvm
 git submodule update --init --recursive
 
+# Apply local patches to the tvm submodule (kept under 3rdparty/patches/).
+# These are minimal fixes we cannot land in the pinned submodule commit, e.g.
+# dynamic-slice support in Buffer.__getitem__ (issue #1207). The shared
+# apply_tvm_patches.sh is also used by build_wheel_ascend.sh and setup.py so
+# that every build path (including CI and `pip install -e .`) picks them up
+# right after the submodule checkout.
+#
+# Behaviour (see 3rdparty/patches/apply_tvm_patches.sh):
+#   - idempotent: an already-applied patch is detected (reverse --check) and
+#     skipped, so re-running install / incremental builds is safe;
+#   - FATAL on failure: if a patch cannot apply (e.g. the pinned tvm was bumped
+#     and the context no longer matches) we exit non-zero instead of silently
+#     building an unpatched TVM.
+bash 3rdparty/patches/apply_tvm_patches.sh
+
 # 根据增量编译选项决定是否清理 build 目录
 if $INCREMENTAL_BUILD; then
     if [ -d build ]; then
--- a/src/target/codegen_ascend_pto.cc
+++ b/src/target/codegen_ascend_pto.cc
@@ -1609,7 +1609,77 @@ void CodeGenTileLangAscendPto::FillCodegen(const CallNode *op) {
   this->PrintIndent();
   this->stream << "wait_flag(PIPE_V, PIPE_S, EVENT_ID0);\n";
 
-  ShapeInfo dst_shape_info = GetSliceInfo(op->args[1].as<CallNode>());
+  const CallNode *dst_access = op->args[1].as<CallNode>();
+  ShapeInfo dst_shape_info = GetSliceInfo(dst_access);
+
+  // A runtime-dynamic slice length (e.g. T.tile.fill(buf[0, 0:idx], v))
+  // cannot be baked into the tile's static valid dims. For such an access
+  // GetSliceInfo falls back to the full tile shape, which makes TEXPANDS
+  // over-fill: it fills the descriptor's *valid* region (there is no
+  // separate count argument). TEXPANDS fills a [valid_row, valid_col]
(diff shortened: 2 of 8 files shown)


src/target/codegen_ascend_pto.cc
```diff
@@ -1609,7 +1609,41 @@ void CodeGenTileLangAscendPto::FillCodegen(const CallNode *op) {
...
+  // over-fill: it fills the descriptor's *valid* region (there is no
+  // separate count argument). Emit a pto::DYNAMIC descriptor whose valid
+  // rows/cols are computed from the dynamic length at runtime, so exactly
+  // the requested elements are filled.
+  if (dst_access->args[3].as<IntImmNode>() == nullptr) {
+    const std::string type = dst_shape_info.type;
+    const std::string col = std::to_string(dst_shape_info.col);
+    const std::string ext = "(" + PrintExpr(dst_access->args[3]) + ")";
+    const std::string valid_row =
+        "(" + ext + " / " + col + " > 1 ? " + ext + " / " + col + " : 1)";
```
LLMZhangYC: With C++ integer division, the formula silently drops tail elements when ext spans multiple rows but isn't a column-aligned multiple. For ext=48, col=32: fills only 32 elements, dropping 16.

testing/python/language/test_tilelang_ascend_language_dynamic_fill.py
```diff
@@ -0,0 +1,89 @@
...
+    ):
+        with T.Kernel(m_num * n_num, is_npu=True) as (cid, vid):
+            bx = cid // n_num
+            by = cid % n_num
+
+            c_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
+            # Zero the tile so the region past idx is a known 0.
+            T.tile.fill(c_ub, 0.0)
+            # Runtime-dynamic fill length: 32 or 64 depending on a value in A.
+            idx = T.if_then_else(A[0, 0] > 0, 32, 64)
```
LLMZhangYC, line 54: Current test has idx ≤ 64 < block_N(256), so fills always stay within one row. Add a case where idx could span multiple rows to guard against the formula bug.

install_ascend.sh
```diff
@@ -173,6 +173,41 @@ echo "Cloning TVM repository and initializing submodules..."
...
+TVM_PATCH_DIR="3rdparty/patches"
+if [ -d "$TVM_PATCH_DIR" ]; then
+    for patch in "$TVM_PATCH_DIR"/tvm_*.patch; do
+        [ -e "$patch" ] || continue
+        patch_name="$(basename "$patch")"
+        patch_abs="$(cd "$(dirname "$patch")" && pwd)/$patch_name"
+        if git -C 3rdparty/tvm apply --reverse --check "$patch_abs" >/dev/null 2>&1; then
+            echo "  [patch] $patch_name already applied, skipping"
+        elif git -C 3rdparty/tvm apply --check "$patch_abs" >/dev/null 2>&1; then
+            git -C 3rdparty/tvm apply "$patch_abs"
```
liggest: Does this patching method support installation using `bash build_wheel_ascend.sh` or `USE_ASCEND=true pip install -e .`?

ChaoyangJi:
use [BufFix] instead of [Fix] in the PR title prefix.

LLMZhangYC (approved):
lgtm

src/target/codegen_ascend_pto.cc
fuhouyu-hw: done

testing/python/language/test_tilelang_ascend_language_dynamic_fill.py
fuhouyu-hw: done

install_ascend.sh
fuhouyu-hw: done

ChaoyangJi (approved):
approve

PR #1309, opened 2026-07-07
[bugfix] Ascend codegen: emit integer max/min as a ternary

### LINKED ISSUE #475: [Bug] `T.min` and `T.max` cause compilation failure when being used with `cid`
Take the following sample code as example:

```python
import tilelang
import tilelang.language as T

@tilelang.jit(out_idx=[-1])
def test():

    @T.prim_func
    def main(out: T.Tensor([64], "float")):
        with T.Kernel(64, is_npu=True) as (cid, vid):
            out[cid] = T.min(cid, 32)
            # out[cid] = T.max(cid, 32)    # also fails!
 
    return main

kernel = test()
print(kernel())
```

Which causes:
```python
/tmp/tmp4jm8dpkl.cpp:29:29: error: call to 'min' is ambiguous
Fatal Err: A fatal error occurred in iternel process, and Diagnostic info can not be reported correctly
/tmp/tmp4jm8dpkl.cpp:29:29: error: call to 'min' is ambiguous
Fatal Err: A fatal error occurred in iternel process, and Diagnostic info can not be reported correctly
Traceback (most recent call last):

...

File "/data/tilelang-ascend/tilelang/jit/adapter/libgen.py", line 128, in compile_lib
    raise RuntimeError(f"Compilation Failed! {command}")
RuntimeError: Compilation Failed! ['bisheng', '--npu-arch=dav-2201', '-O2', '-std=c++17', '-xasc', '-I/usr/local/Ascend/cann-8.5.0/include', '-I/usr/local/Ascend/cann-8.5.0/include/experiment/msprof', '-I/usr/local/Ascend/cann-8.5.0/include/experiment/runtime', '-I/usr/local/Ascend/cann-8.5.0/pkg_inc', '-I/usr/local/Ascend/cann-8.5.0/pkg_inc/runtime', '-I/usr/local/Ascend/cann-8.5.0/pkg_inc/profiling', '-I/data/tilelang-ascend/3rdparty/catlass/include', '-I/data/tilelang-ascend/3rdparty/shmem/include', '-I/data/tilelang-ascend/3rdparty/shmem/src/device', '-DBACKEND_HYBM', '-I/data/tilelang-ascend/tilelang/../src', '-L/usr/local/Ascend/cann-8.5.0/lib64', '-Wno-macro-redefined', '-Wno-ignored-attributes', '-Wno-non-c-typedef-for-linkage', '-lruntime', '-lascendcl', '-lm', '-ltiling_api', '-lplatform', '-lc_sec', '-ldl', '-fPIC', '--shared', '-g', '/tmp/tmp4jm8dpkl.cpp', '-o', '/tmp/tmp4jm8dpkl.so']
```

yangqiang2018:
Fixes #475

## What

`CodeGenTileLangAscend` (the non-PTO Ascend backend) inherited `CodeGenC`'s default emission for scalar `Max`/`Min` nodes — a bare `max(a, b)` / `min(a, b)`. In the Ascend (bisheng/CANN) C++ environment those unqualified names are ambiguous for mismatched integer widths (e.g. `max(int64_t, int)` from an int64 grid variable and an int literal), so any kernel that emits a **runtime** (non-constant-foldable) scalar integer max/min fails to compile with `error: call to 'max' is ambiguous`. Existing kernels only used compile-time-foldable `T.max` / `T.min`, so the gap stayed latent.

## Fix

Override `VisitExpr_(MaxNode)` / `VisitExpr_(MinNode)` to emit a ternary `(a > b ? a : b)` / `(a < b ? a : b)` for integer/uint dtypes, which needs no overload resolution. Float and all other dtypes fall through to the unmodified `CodeGenC` path, so NaN/float semantics are unchanged and the change is strictly additive for existing operators.

## Test

Adds `testing/python/language/test_tilelang_ascend_language_integer_minmax_index.py`: a kernel that forces a runtime integer `max` / `min` on the grid-block index and uses it as a DMA row offset (so it survives DCE and reaches codegen), across both ops and both `ascendc` / `pto` targets. The `ascendc` cases are the ones that trigger and validate the fix — they fail to compile with the ambiguous-call error without it; the `pto` cases are a control (that path already used `std::max` / `std::min` and is unaffected).

## Validation (Ascend NPU)

New test 4 passed; `testing/python` pytest 827 passed / 0 failed (2 xfail); `examples/bench_test.sh` passes except one pre-existing flaky autotune example (`convolution/example_convolution_autotune.py`, passes standalone, uses no integer max/min).

### DIFF
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -425,6 +425,47 @@ void CodeGenTileLangAscend::VisitExpr_(const FloorModNode *op,
   os << ")";
 }
 
+// Emit an INTEGER max/min as a ternary instead of the bare max(a, b) / min(a,
+// b) CodeGenC produces: in the Ascend (bisheng/CANN) C++ environment those
+// unqualified names are ambiguous for mixed integer widths (e.g. max(int64_t,
+// int) from an int64 loop var and an int literal), and a ternary needs no
+// overload resolution. We avoid std::max<T> because the generated AscendC
+// kernel prelude does not include <algorithm>. Shared by MaxNode (">") and
+// MinNode ("<").
+void CodeGenTileLangAscend::PrintIntMinMaxTernary(const PrimExpr &a,
+                                                  const PrimExpr &b,
+                                                  const char *cmp,
+                                                  std::ostream &os) {
+  os << "(";
+  PrintExpr(a, os);
+  os << " " << cmp << " ";
+  PrintExpr(b, os);
+  os << " ? ";
+  PrintExpr(a, os);
+  os << " : ";
+  PrintExpr(b, os);
+  os << ")";
+}
+
+// Scope to SCALAR integer/uint (lanes == 1) so float and vector max/min keep
+// CodeGenC's default (unchanged NaN / vector semantics), keeping the change
+// strictly additive for everything else.
+void CodeGenTileLangAscend::VisitExpr_(const MaxNode *op, std::ostream &os) {
+  if ((op->dtype.is_int() || op->dtype.is_uint()) && op->dtype.lanes() == 1) {
+    PrintIntMinMaxTernary(op->a, op->b, ">", os);
+  } else {
+    CodeGenC::VisitExpr_(op, os);
+  }
+}
+
+void CodeGenTileLangAscend::VisitExpr_(const MinNode *op, std::ostream &os) {
+  if ((op->dtype.is_int() || op->dtype.is_uint()) && op->dtype.lanes() == 1) {
+    PrintIntMinMaxTernary(op->a, op->b, "<", os);
+  } else {
+    CodeGenC::VisitExpr_(op, os);
+  }
+}
+
 void CodeGenTileLangAscend::VisitExpr_(const BufferLoadNode *op,
                                        std::ostream &os) {
   auto var_name = var_idmap_[op->buffer->data.get()];
--- a/src/target/codegen_ascend.h
+++ b/src/target/codegen_ascend.h
@@ -51,6 +51,10 @@ public:
   void VisitExpr_(const CallNode *op, std::ostream &os) final;
   void VisitExpr_(const FloorDivNode *op, std::ostream &os);
   void VisitExpr_(const FloorModNode *op, std::ostream &os);
+  void VisitExpr_(const MaxNode *op, std::ostream &os);
+  void VisitExpr_(const MinNode *op, std::ostream &os);
+  void PrintIntMinMaxTernary(const PrimExpr &a, const PrimExpr &b,
+                             const char *cmp, std::ostream &os);
   void VisitExpr_(const MulNode *op, std::ostream &os) final;
   void VisitExpr_(const SelectNode *op, std::ostream &os) final;
   void VisitExpr_(const BufferLoadNode *op, std::ostream &os) final;
--- a/testing/python/language/test_tilelang_ascend_language_integer_minmax_index.py
+++ b/testing/python/language/test_tilelang_ascend_language_integer_minmax_index.py
@@ -1,0 +1,111 @@
+import pytest
+
+import torch
+
+import tilelang
+import tilelang.language as T
+
+"""
+Regression test for integer scalar ``max`` / ``min`` in the AscendC codegen.
+
+A kernel that uses a *runtime* (non-constant-foldable) scalar ``T.max`` / ``T.min``
+on a grid-block index lowers to a ``Max`` / ``Min`` IR node whose operands have
+mismatched integer widths -- an int64 block variable against an int literal.
+
+The non-PTO ``CodeGenTileLangAscend`` backend inherits ``CodeGenC``'s default and
+prints these as a bare ``max(a, b)`` / ``min(a, b)``.  bisheng then rejects the
+generated kernel with ``error: call to 'max' is ambiguous`` (the int64/int
+overloads are equally viable), so the kernel fails to compile.  Emitting the
+integer case as a ternary ``(a > b ? a : b)`` / ``(a < b ? a : b)`` removes the
+overload resolution entirely.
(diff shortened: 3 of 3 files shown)


yangqiang2018:
Thanks for the review!

I kept the ternary here instead of `std::max<T>` / `std::min<T>` on purpose:

1. **Self-contained.** The ternary lowers to plain operators, so it needs no headers in the generated kernel (the emitted code is what `bisheng` compiles). `std::max<T>` / `std::min<T>` would instead rely on `<algorithm>` being available in the generated Ascend kernel.
2. **Bounded in practice.** The operands that reach this path are simple, non-nested index/boundary expressions (e.g. `max(s - w, 0)`, `min(a, b)`), so the duplication is a couple of extra subexpressions rather than the exponential growth that deeply nested max/min could cause.

If you'd prefer `std::max<T>` / `std::min<T>`, I'm happy to switch — it's a small change on top of ensuring `<algorithm>` is available in the generated kernel.

yangqiang2018:
The `Benchmark Tests (Ascend NPU)` failure here is a pre-existing flaky test, unrelated to this change — could a maintainer re-run the job?

The failing test is `testing/python/language/parallel/test_tilelang_ascend_language_parallel.py::TestTileLangKernels::test_shiftright_operation`. It chooses the shift amount with `scalar_value = random.randint(1, 32)`, and `random.randint` is inclusive, so it can pick `32`. Shifting an `int32` by 32 then trips the compile-time check:

```
InternalError: Check failed: (pb->value >= 0 && pb->value < rtype.bits()) is false:
Shift amount must be non-negative and less than 32 for type int32
```

The `setup_random_seed` fixture only calls `torch.manual_seed(0)`, not `random.seed(...)`, so the shift amount is effectively unseeded and this fails intermittently (~1/32 of runs).

It's independent of this PR: the change here only touches integer `max`/`min` codegen, the failing test is a right shift with no `max`/`min`, the error is raised at the IR-verification stage (before codegen), and this same commit passed the full `testing/python` suite (827 passed, 2 xfailed) plus all 137 examples on my Ascend NPU. A re-run should be green.

(Happy to send a separate PR narrowing the range to `random.randint(1, 31)` if that's useful.)

fuhouyu-hw (dismissed):
lgtm

src/target/codegen_ascend.cc
```diff
@@ -425,6 +425,45 @@ void CodeGenTileLangAscend::VisitExpr_(const FloorModNode *op,
...
+    PrintExpr(op->a, os);
+    os << " : ";
+    PrintExpr(op->b, os);
+    os << ")";
+  } else {
+    CodeGenC::VisitExpr_(op, os);
+  }
+}
+
+void CodeGenTileLangAscend::VisitExpr_(const MinNode *op, std::ostream &os) {
```
erhsh, line 461: 问题：大量重复模板代码：Min/Max 逻辑高度对称
建议：抽取公共辅助函数消除重复逻辑

yangqiang2018: 说得对 —— 已抽出公共 helper `PrintIntMinMaxTernary`,`VisitExpr_(MaxNode)`(传 `>`)和 `VisitExpr_(MinNode)`(传 `<`)都调它,int/uint 判断和三元输出不再重复,生成代码逐字节不变。

（保留三元、没用 `std::max<T>` / `std::min<T>`：生成的 AscendC kernel 前言里没有 `<algorithm>`，用 `std::` 还得额外加 include；pto 路径那个 `std::max` 是 host 侧 codegen，上下文不同。）

src/target/codegen_ascend.cc
```diff
@@ -425,6 +425,46 @@ void CodeGenTileLangAscend::VisitExpr_(const FloorModNode *op,
...
+  PrintExpr(a, os);
+  os << " : ";
+  PrintExpr(b, os);
+  os << ")";
+}
+
+// Scope to integer/uint so float max/min keep CodeGenC's default (unchanged
+// NaN semantics), keeping the change strictly additive for non-integer ops.
+void CodeGenTileLangAscend::VisitExpr_(const MaxNode *op, std::ostream &os) {
+  if (op->dtype.is_int() || op->dtype.is_uint()) {
```
ChaoyangJi: 这里只针对标量类型吧，最好增加个数据宽度的保护:
if ((op->dtype.is_int() || op->dtype.is_uint()) && op->dtype.lanes() == 1)

liggest:
Thanks for your patch! 
There's actually an relevant issue about this fix: https://github.com/tile-ai/tilelang-ascend/issues/475
Would you mind to also check the sample code described in the issue? It's likely to be fixed by this patch as well

src/target/codegen_ascend.cc
yangqiang2018: 好的，已加上 `&& op->dtype.lanes() == 1`，现在只对**标量**整数 max/min 用三元；向量整数（以及浮点）继续走 CodeGenC 的默认路径。已 push。

yangqiang2018:
@liggest Thanks! Yes — this patch fixes exactly that. #475's `T.min(cid, 32)` / `T.max(cid, 32)` produces the same ambiguous `min(int64_t, int)` / `max(int64_t, int)` in the generated AscendC kernel, and the regression test added here covers the same pattern (a runtime integer max/min on the grid-block index). I've added `Fixes #475` to the PR description so it closes on merge.

ChaoyangJi (approved):
approve

ChaoyangJi:
using [bugfix] instead of [Fix] in the PR titile.

yangqiang2018:
@ChaoyangJi Done — updated the title to `[bugfix]`. Thanks for the review and the approval!

erhsh:
lgtm

ChaoyangJi (approved):
approve

fuhouyu-hw (approved):
approve

PR #1312, opened 2026-07-07
[Bugfix] Ascend gemm_v0: N-tiling to fit L0B for large N

yangqiang2018:
## What

`gemm_v0` (the AscendC / non-PTO matmul template in `src/tl_templates/ascend/common.h`) tiled only the K dimension: every `mma` loaded the whole N of the B operand into a single L0B slot (`N * kL0Size`). L0B is 64KB and the kL0 ping-pong halves the per-slot budget to 32KB, so once N is large the B sub-tile overflows L0B and the cube crashes at run time.

Concretely an attention PV matmul is `<M = 64, N = headDim = 512, K = window <= 128>`: the B tile is `128 * 512 * 2 = 128KB`, twice the whole L0B.

## Fix

Add an inner N-tiling loop (`nTile = 128`, the largest N-tile whose `(kL0Size x nTile)` B sub-block fits the 32KB per-ping-pong-slot L0B budget): each N-tile loads its own `(kL0Size x nTile)` B sub-block into L0B and accumulates into its own column band of the L0C output.

The `(N-tile, K-tile)` sequence is flattened and the L0A/L0B ping-pong is primed/drained **once**, running continuously across all tiles, so each tile's L1->L0 load overlaps the previous tile's `mma` (no per-tile serialisation).

Sub-tile offsets come straight from the catlass `tla` fractal layouts (verified against `3rdparty/catlass`, not guessed):
- L0C column `n0` → `n0 * roundUp16(M)` (`tla::MakeLayoutL0C` N1 stride)
- L1 zN B column `n0` → `n0 * roundUp16(K)` (`tla::MakeLayout<zN>` C1 stride)

both consistent with the K-offset `B[kL0Idx*16*kL0Size]` the original loop already used.

## Compatibility

Only `transpose_B=false` with `N > nTile` is tiled. Every existing caller (`transpose_B`, or `N <= nTile`) gets `nL0split == 1`, `tileIdx == kL0Idx`, and **byte-identical codegen**, so no working caller changes. Large-N `transpose_B=false` previously overflowed L0B, so this only adds a path that could not run before. A `static_assert` guards non-divisible N.

## Test

`testing/python/language/test_tilelang_ascend_language_gemm_v0_n_tiling.py` — a plain `C = A @ B` (`transpose_B=false`) staged through L1, checked against torch:
- `N = 512` (the PV shape) — the large-N path that overflowed L0B before this fix. `K = 128` (single K-tile, N tiled into 4) and `K = 256` (flattened 2 K-tiles × 4 N-tiles pipeline).
- `N = 128` (`<= nTile`) — the compatibility case: `nL0split == 1`, original single-pass path.

Verified on Ascend NPU: the new test passes (6/6), and the existing `gemm_v0` suite (`test_tilelang_ascend_language_gm_to_l1.py`, 49 cases) still passes.

## Note on the PTO backend

This targets the **ascendc** backend only. The PTO backend has its own separate `gemm_v0` (`src/tl_templates/pto/common.h`, a different `gemm_v0_inner` / `TileMatL0B<..., N, ...>` implementation) that allocates the whole N in one L0B tile as well, so it has the same large-N L0B-overflow limitation. It is a different abstraction and needs a different change, tracked separately in #1313.

### DIFF
--- a/src/tl_templates/ascend/common.h
+++ b/src/tl_templates/ascend/common.h
@@ -602,6 +602,67 @@ gemm_v0(LocalTensor<T1> const &A, LocalTensor<T1> const &B,
   uint32_t kL0Tail = K - (kL0split - 1) * kL0Size;
   bool initflag = false;
 
+  // ---- N tiling -----------------------------------------------------------
+  // The B operand tile loaded into L0B is (kL0Size x nTile); L0B holds 64KB,
+  // and with the kL0 ping-pong the per-slot budget is 32KB. So a single mma
+  // over the whole N (l0b slot = N*kL0Size) overflows L0B once N is large
+  // (e.g. the PV matmul's N = headDim = 512 -> 512*128*2 = 128KB). Tile N into
+  // nTile columns just like the Ascend C reference (N_SPLIT_SIZE = 128): each
+  // tile loads its own (kL0Size x nTile) B sub-block and writes its own column
+  // band of the L0C accumulator. Only the non-transpose-B path is tiled; the
+  // transpose-B callers (e.g. QK with N = block_I <= 128) already fit, so they
+  // keep nTile == N (a single pass, byte-for-byte the original behaviour) and
+  // need no L1 column-offset formula. Compatibility: any existing caller with
+  // N <= nMaxByL0B (transpose or not) sees nL0split == 1 and identical codegen.
+  //
+  // Sub-tile offsets come straight from the catlass tla fractal layouts:
+  //   L0C column n0  ->  n0 * roundUp16(M)   (tla::MakeLayoutL0C N1 stride)
+  //   L1 zN B col n0 ->  n0 * roundUp16(K)   (tla::MakeLayout<zN>  C1 stride)
+  // both of which are consistent with the original K-offset
+  // B[kL0Idx*16*kL0Size] (zN K-row stride) already used below.
+  constexpr uint32_t nMaxByL0B = (32u * 1024u) / (kL0Size * sizeof(T1));
+  constexpr uint32_t nTile = (transpose_B || N <= nMaxByL0B) ? N : nMaxByL0B;
+  static_assert(transpose_B || (N % nTile == 0),
+                "gemm_v0 N-tiling requires N divisible by the N tile size");
+  constexpr uint32_t nL0split = N / nTile;
+  // L0A and L0B are each 64KB. The kL0/N ping-pong uses two slots only when
+  // there is more than one (N-tile, K-tile) step; a single step uses one slot
+  // and may occupy the whole 64KB -- which is why a transpose-B caller with a
+  // single K-tile and N up to 64KB/(kL0Size*sizeof(T)) fits (e.g. fp32 N=128 =
+  // 64KB). So the per-slot budget is 64KB for a single step, 32KB once the
+  // ping-pong actually alternates.
+  constexpr uint32_t kNumSteps = ((K + kL0Size - 1) / kL0Size) * nL0split;
+  constexpr uint32_t kL0Budget = (64u * 1024u) / (kNumSteps > 1 ? 2u : 1u);
+  // The B tile in L0B is (kL0Size x nTile) -- only the transpose-B path keeps
+  // nTile == N, so a large-N transpose-B caller could overflow its L0B slot.
+  static_assert(nTile * kL0Size * sizeof(T1) <= kL0Budget,
+                "gemm_v0: the (kL0Size x nTile) B tile does not fit its L0B "
+                "ping-pong slot");
+  // M is not tiled, so a large M would overflow its L0A slot.
+  static_assert(M * kL0Size * sizeof(T1) <= kL0Budget,
+                "gemm_v0: the (M x kL0Size) A tile does not fit its L0A "
+                "ping-pong slot");
+  // L0C is caller-allocated (the `C` operand) and holds the full (M x N)
+  // accumulator: roundUp16(M) * N * sizeof(T2). It must fit the target's L0C
+  // (A2/A3 128KB, A5 256KB) -- e.g. M=128, N=512, fp32 needs 256KB and only
+  // fits A5. Unlike the L0A/L0B guards above this can't be a static_assert
+  // here: L0C capacity is device-dependent, whereas L0A/L0B are 64KB on both
+  // archs (which is why those two use a literal budget). This device-side
+  // template has no arch macro or L0C-size constant to branch on
+  // (ASCEND_*_L0C_SIZE live in the host codegen), so a literal guard would
+  // reject a valid A5 caller or silently pass on A2/A3. The constraint is
+  // therefore documented here; the caller sizes its L0C tile accordingly.
+  constexpr uint32_t mRound = ((M + 15u) / 16u) * 16u;
+  constexpr uint32_t kRound = ((K + 15u) / 16u) * 16u;
+
+  // ---- Pipelined main loop. Prime/drain the L0A/L0B ping-pong buffers ONCE
+  // and let the ping-pong run continuously across the WHOLE (N-tile, K-tile)
+  // sequence (flattened by tileIdx). Each tile's L1->L0 load goes into the
+  // free buffer while the previous tile's mma runs, so N-tiles overlap with K
+  // exactly like the Ascend C matmul pipeline -- a per-N-tile drain (the first
+  // N-tiling version) instead serialised the tiles. For nL0split == 1 (every
+  // pre-existing caller, and the QK transpose-B path) tileIdx == kL0Idx, so
+  // this is byte-for-byte the original K ping-pong.
   SetFlag<HardEvent::MTE2_MTE1>(L0AB_EVENT);
   WaitFlag<HardEvent::MTE2_MTE1>(L0AB_EVENT);
   SetFlag<HardEvent::FIX_M>(L0AB_EVENT);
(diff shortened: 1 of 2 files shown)


src/tl_templates/ascend/common.h
```diff
@@ -610,35 +644,44 @@ gemm_v0(LocalTensor<T1> const &A, LocalTensor<T1> const &B,
...
-    }
-    if constexpr (!transpose_B) {
-      tl::ascend::copy_l1_to_l0b<T1, K, N>(l0b[l0b_base],
-                                           B[kL0Idx * 16 * kL0Size], kSize, N);
-    } else {
-      tl::ascend::copy_l1_to_l0b<T1, N, K, true>(
-          l0b[l0b_base], B[kL0Idx * N * kL0Size], kSize, N);
+  uint32_t tileIdx = 0;
+  for (uint32_t nL0Idx = 0; nL0Idx < nL0split; nL0Idx++) {
+    uint32_t bNOffset = transpose_B ? 0u : (nL0Idx * nTile * kRound);
```
LLMZhangYC, line 680: Correctness depends entirely on B_L1 being allocated in zN layout. Add a static_assert on layout type or at minimum a prominent comment.

testing/python/language/test_tilelang_ascend_language_gemm_v0_n_tiling.py
```diff
@@ -0,0 +1,125 @@
...
+    torch.testing.assert_close(c, ref, rtol=rtol, atol=atol)
+
+
+# (M, N, K) -- N=512 is the large-N path that overflowed L0B before N-tiling;
+# N=128 (<= nTile) is the single-pass compatibility case.  M=64 keeps the N=512
+# L0C tile (M*N*4) within the 128KB L0C budget, matching the real PV matmul.
+gemm_configs = [
+    (64, 512, 128),  # PV shape: single K-tile, N tiled into 4
+    (64, 512, 256),  # flattened (N-tile, K-tile) pipeline: 2 K-tiles x 4 N-tiles
+    (128, 128, 128),  # N <= nTile: nL0split == 1, original single-pass path
```
LLMZhangYC, line 114: Add (64, 512, 200): kL0split=2, kL0Tail=72, nL0split=4 — tests K-tailing under N-tiling simultaneously.

src/tl_templates/ascend/common.h
```diff
@@ -602,6 +602,40 @@ gemm_v0(LocalTensor<T1> const &A, LocalTensor<T1> const &B,
...
+  // need no L1 column-offset formula. Compatibility: any existing caller with
+  // N <= nMaxByL0B (transpose or not) sees nL0split == 1 and identical codegen.
+  //
+  // Sub-tile offsets come straight from the catlass tla fractal layouts:
+  //   L0C column n0  ->  n0 * roundUp16(M)   (tla::MakeLayoutL0C N1 stride)
+  //   L1 zN B col n0 ->  n0 * roundUp16(K)   (tla::MakeLayout<zN>  C1 stride)
+  // both of which are consistent with the original K-offset
+  // B[kL0Idx*16*kL0Size] (zN K-row stride) already used below.
+  constexpr uint32_t nMaxByL0B = (32u * 1024u) / (kL0Size * sizeof(T1));
+  constexpr uint32_t nTile = (transpose_B || N <= nMaxByL0B) ? N : nMaxByL0B;
```
LLMZhangYC, line 624: l0a ping-pong slot size M * kL0Size * sizeof(T1) has no capacity guard. For large M (e.g. M=512), it overflows 64KB L0A silently. Add a static_assert.

ChaoyangJi:
use [Bugfix] instead of [Fix] prefix in the PR tile.

src/tl_templates/ascend/common.h
```diff
@@ -602,6 +602,40 @@ gemm_v0(LocalTensor<T1> const &A, LocalTensor<T1> const &B,
...
   bool initflag = false;
 
+  // ---- N tiling -----------------------------------------------------------
+  // The B operand tile loaded into L0B is (kL0Size x nTile); L0B holds 64KB,
+  // and with the kL0 ping-pong the per-slot budget is 32KB. So a single mma
+  // over the whole N (l0b slot = N*kL0Size) overflows L0B once N is large
+  // (e.g. the PV matmul's N = headDim = 512 -> 512*128*2 = 128KB). Tile N into
+  // nTile columns just like the Ascend C reference (N_SPLIT_SIZE = 128): each
+  // tile loads its own (kL0Size x nTile) B sub-block and writes its own column
+  // band of the L0C accumulator. Only the non-transpose-B path is tiled; the
```
benyang0506, line 612: Thanks for the clarification. I think tiling n should be independent of whether B is transposed, and ideally both paths should support it.

If transpose-B N-splitting is not supported in this PR, could we add a compile-time guard to make the assumption explicit? Otherwise a future transpose-B caller with larger N may silently overflow L0B.

Like:
```cpp  
static_assert(!transpose_B || N <= nMaxByL0B,  
              "transpose_B gemm_v0 requires N to fit in L0B");  
```

yangqiang2018: Added a guard as you suggested. The first cut used a flat `N <= nMaxByL0B` (a 32KB-per-slot budget), but that rejected a valid existing caller — `examples/gemv/example_gemv_c.py` runs a transpose_B `gemm_v0` with fp32 N=128, which is a single `(N-tile, K-tile)` step and legitimately uses one full 64KB L0B slot (64KB fits). So the budget is now ping-pong-aware: 64KB for a single step, 32KB once the ping-pong actually alternates (`kL0split * nL0split > 1`). It still catches a genuinely oversized `nTile`/`M` (e.g. a large-N transpose_B caller, which is what you flagged), it just no longer trips the single-step callers. gemv compiles and passes again on the updated branch — your point stands, the budget just needed to be step-count-aware.

src/tl_templates/ascend/common.h
yangqiang2018: Good catch. Added a guard so an oversized M fails at compile time instead of silently overflowing its L0A slot. It shares the same ping-pong-aware budget as the L0B guard: 64KB for a single `(N-tile, K-tile)` step, 32KB once the ping-pong alternates (`kL0split * nL0split > 1`), so a legitimate single-step caller that uses a full 64KB slot isn't rejected.

src/tl_templates/ascend/common.h
yangqiang2018: gemm_v0 takes a raw LocalTensor with no layout type, so a static_assert on the B layout is not reachable here. I added a prominent comment at the bNOffset computation instead (312b3005): it documents that the column offset nL0Idx*nTile*kRound is the zN C1 stride and is correct only for a zN B_L1, so a different B layout would need a different per-column stride.

testing/python/language/test_tilelang_ascend_language_gemm_v0_n_tiling.py
yangqiang2018: I first added `(64, 512, 200)` to test a K-tail under N-tiling, but it turned out to expose a **pre-existing** `gemm_v0` bug unrelated to this PR: the unaligned K-tail (`kL0Tail = 72`) is miscomputed even with no N-tiling — `(64, 128, 200)` (N=128, so `nL0split == 1`, `bNOffset == 0`) fails identically at 10.47%, as does the parent commit `1319a0da` (before any change in this PR), and for `nL0split == 1` this PR's K-tile inner loop is byte-identical to the original (`tileIdx == kL0Idx`, `bNOffset == 0`). So I've dropped `(64, 512, 200)` and filed #1341 to track it separately; the aligned-K path under N-tiling stays covered by `(64, 512, 256)` (two 16-aligned K-tiles). Thanks — it surfaced a real bug even though it's outside this PR's scope.

src/tl_templates/ascend/common.h
```diff
@@ -602,6 +602,57 @@ gemm_v0(LocalTensor<T1> const &A, LocalTensor<T1> const &B,
...
+  // ping-pong actually alternates.
+  constexpr uint32_t kNumSteps = ((K + kL0Size - 1) / kL0Size) * nL0split;
+  constexpr uint32_t kL0Budget = (64u * 1024u) / (kNumSteps > 1 ? 2u : 1u);
+  // The B tile in L0B is (kL0Size x nTile) -- only the transpose-B path keeps
+  // nTile == N, so a large-N transpose-B caller could overflow its L0B slot.
+  static_assert(nTile * kL0Size * sizeof(T1) <= kL0Budget,
+                "gemm_v0: the (kL0Size x nTile) B tile does not fit its L0B "
+                "ping-pong slot");
+  // M is not tiled, so a large M would overflow its L0A slot.
+  static_assert(M * kL0Size * sizeof(T1) <= kL0Budget,
```
fuhouyu-hw, line 642: L0A and L0B overflow are guarded by static_assert, but L0C is not ,on A2/A3 with only 128KB L0C, an M=128, N=512, fp32 case silently overflows; please add a capacity guard or at minimum document the constraint.

LLMZhangYC (dismissed):
lgtm

src/tl_templates/ascend/common.h
yangqiang2018: Good point — L0C is unguarded. I've documented the constraint (`roundUp16(M) * N * sizeof(T2)` must fit L0C) at the guard block rather than static_assert it, for a concrete reason: L0C capacity is device-dependent (128KB on A2/A3, 256KB on A5), whereas L0A/L0B are 64KB on both — that's why those two guards can use a literal budget. This device-side template has no arch macro or L0C-size constant to branch on (`ASCEND_*_L0C_SIZE` live in the host `codegen_ascend.cc`, not in `common.h`), so a literal L0C static_assert would either reject a legitimate A5 caller (128KB < M*N*4 ≤ 256KB) or silently pass on A2/A3. Same situation as the B-layout note — asserted where reachable, documented where not. Your M=128 / N=512 / fp32 case (256KB) is exactly the A5-only scenario the new comment calls out.

benyang0506 (approved):
lgtm

ChaoyangJi (approved):
approve

PR #1330, opened 2026-07-09
[Feat] Ascend: add copy_pa primitive for paged KV load

yangqiang2018:
## What

Add a `copy_pa` primitive (`T.copy_pa`) that loads a window of a **paged** KV cache straight from GM into an L1 tile, resolving each page through a `block_table`. The window is walked one page at a time, and each page's contiguous run of rows is Nd2Nz-copied into L1, so a window that sits in one page is a single `DataCopy`. Running on the cube (L1-direct) it needs no UB staging and no GM workspace round-trip — it replaces a per-row gather (GM→UB→GM through a workspace) that is the single largest source of the performance gap on paged attention.

## Why this needs a primitive (it can't be composed from existing ones)

The core of a paged KV load is a **data-dependent, per-page `while` loop**:

```c
while (rows_done < window) {
    page     = blockTable[s2Idx / blockSize];        // indirect page lookup
    copyRows = blockSize - (s2Idx % blockSize);      // runtime rows this page
    DataCopy(L1[rows_done], kv[page*kvStride + ...], Nd2Nz(copyRows rows));
    s2Idx += copyRows;  rows_done += copyRows;
}
```

Composing this from existing TileLang primitives fails on **three independent runtime dimensions**, plus a missing operand:

1. **The loop trip count is runtime.** How many pages a window spans depends on `s2Idx` (different per token), but `T.serial(N)` requires a *compile-time* `N`. Even bounding the page count to a compile-time upper bound and fully unrolling does not escape (2) and (3).
2. **The per-page row count is runtime** (`blockSize - s2Idx % blockSize`). `T.copy`'s transfer extent must be compile-time (it feeds the `DataCopy` / `Nd2Nz` parameters), so it cannot accept a runtime row count.
3. **The sub-window offset into the matmul is runtime.** Even if a whole fixed-size page were staged into L1 and the matmul asked to take `[brow : brow+window]`, the start `brow` is runtime — and `gemm_v0` rejects sliced / runtime-offset operands (`Unsupported BufferLoad`).

On top of those, **`T.copy` has no block-table operand at all** — it only selects `copy_gm_to_l1` etc. by scope, so paged indirect addressing simply cannot be expressed with it.

The only "kernel-side" alternative — a fixed-size batch copy that assumes the window never crosses a page — holds only in a degenerate config (where `blockSize` is large enough that the window happens to fit one page), still needs the runtime `brow` slice of (3), and does not model the paged dataflow (which loops per page and may cross pages).

## Generality (not specific to one op)

A paged KV load is the **general** paged-attention data path: a paged KV cache + block table is the standard PagedAttention layout, and "read a paged window straight into L1" is what any paged-attention cube side must do. Wrapping a C template as a primitive **is** how the TileLang Ascend backend already works — `gemm_v0`, `copy_gm_to_l1`, `copy_l1_to_l0a`, every `T.tile.*` is a `common.h` template exposed via builtin + codegen + Python binding. `copy_pa` is isomorphic to them.

## Compatibility

Pure addition: a new op (`ascend_copy_pa`), a new `common.h` template, a new codegen branch, two config-table rows, and the `T.copy_pa` Python binding. No existing op, template, or codegen path changes.

## Test

`testing/python/language/test_tilelang_ascend_language_copy_pa.py` builds a paged KV cache with a **shuffled** block table (physical page ≠ logical page, to exercise the indirection), loads a window via `copy_pa`, and checks a QK matmul consuming the L1 tile against a hand-gathered torch reference. Coverage (ascendc target):

- single page — window fits one page (`while` runs once)
- cross 2 pages — window straddles a page boundary (`while` runs twice)
- cross 3 pages, mid-page start — runtime per-page row count + offset (`while` runs three times)

Verified on Ascend NPU: the new test passes 6/6, and the existing `gemm_v0` suite (`test_tilelang_ascend_language_gm_to_l1.py`, 49 cases) still passes.

## Note on the PTO backend

This targets the **ascendc** backend only. The PTO backend has no paged KV load primitive (there is no `copy_pa` equivalent in `src/tl_templates/pto/common.h` or its codegen); it is a different abstraction and needs a different implementation, tracked separately in #1331.

### DIFF
--- a/src/op/ascend.cc
+++ b/src/op/ascend.cc
@@ -1174,6 +1174,11 @@ TIR_DEFINE_TL_BUILTIN(ascend_gemm_v1)
     .set_attr<TCallEffectKind>("TCallEffectKind",
                                Integer(CallEffectKind::kOpaque));
 
+TIR_DEFINE_TL_BUILTIN(ascend_copy_pa)
+    .set_num_inputs(-1)
+    .set_attr<TCallEffectKind>("TCallEffectKind",
+                               Integer(CallEffectKind::kOpaque));
+
 TIR_DEFINE_TL_BUILTIN(ascend_printf)
     .set_num_inputs(-1)
     .set_attr<TCallEffectKind>("TCallEffectKind",
--- a/src/op/ascend.h
+++ b/src/op/ascend.h
@@ -188,6 +188,8 @@ TVM_DLL const Op &ascend_gemm_v0();
 
 TVM_DLL const Op &ascend_gemm_v1();
 
+TVM_DLL const Op &ascend_copy_pa();
+
 TVM_DLL const Op &ascend_printf();
 
 TVM_DLL const Op &ascend_dump_tensor();
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -638,6 +638,8 @@ void CodeGenTileLangAscend::VisitExpr_(const CallNode *op, std::ostream &os) {
     PrintOpCall(op, "AscendC::SyncAll<false>", {0, 0}, {0, 0});
   } else if (op->op.same_as(tl::ascend_gemm_v0())) {
     GemmOpCodegen(op);
+  } else if (op->op.same_as(tl::ascend_copy_pa())) {
+    CopyPACodegen(op);
   } else if (op->op.same_as(tl::ascend_printf())) {
     PrintfOpCodegen(op, "AscendC::PRINTF");
   } else if (op->op.same_as(tl::ascend_dump_tensor())) {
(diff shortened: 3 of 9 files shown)


yangqiang2018:
Closing this for now.

On a closer look, a paged KV load can be composed from existing primitives for the common block-aligned case — `T.copy(cache[runtime_block_id, fixed-size slice])` resolved through a `block_table`, e.g. `examples/flash_attention/paged_flash_attn_bhsd.py` and `examples/HISA/paged_block_sparse_mqa_attn_expert.py`. The runtime block id indexes the cache's block dimension fine; only the per-copy extent needs to be compile-time.

`copy_pa` only adds value for a non-block-aligned sliding window (a partial first page with a runtime per-page row count), and even there it is a bandwidth optimization over "read whole blocks + mask", not something that cannot be expressed. So the "can't be composed from existing primitives" framing in this PR is not right.

I'd like to re-evaluate whether the same performance can be reached with existing primitives before proposing a compiler change, so closing this. Thanks!

PR #1360, opened 2026-07-14
[Ascend] Vector tail-block scheme: opt-in valid-region compute for un…

fuhouyu-hw:
…ary/binary/scalar (AscendC + PTO)

Automatic UB tail handling for the common elementwise cases, gated behind an explicit pass config so non-tail kernels are completely unaffected.

- New pass config TL_ASCEND_TAIL_MASK (default False). When enabled, the new pass AscendTailMaskPropagation (after LowerTileOp) seeds the valid [row,col] rect from copy_gm_to_ub, propagates it through the UB dataflow, and rewrites unary/binary/scalar on a clean 2D float tail tile to internal tl.ascend_tail_* ops carrying runtime valid_row/valid_col/physical_col.
- Conservative guards (kept): element count == physical tile, float-like dtype, no out-of-scope loop var, not a broadcast scalar. 3D/int/mismatched tiles bail to the full-tile op.
- reduce is NOT rewritten (pad-filled full-tile path); tail_reduce ReduceSum<AR> has an output-layout bug. reduce/compare/select/broadcast tail is future work.
- pad_value / copy_gm_to_ub gap-fill kept so full-tile readers see a defined value; softmax uses pad_value=-inf and does not enable the flag.
- Codegen: AscendC tail_{unary,binary,scalar} helpers; PTO tail_{unary,binary, scalar} over a TileUbDataND<...DYNAMIC> tile. Min/Max emitted as ternaries; valid-region exprs bound before the call line.

Rebased onto upstream/ascendc_pto (merged with brcb/row_expand ops).

### DIFF
--- a/examples/tail_mask/example_tail_add.py
+++ b/examples/tail_mask/example_tail_add.py
@@ -1,0 +1,65 @@
+"""Minimal tail-block elementwise-add example for the AscendC backend.
+
+M and N are deliberately *not* multiples of the block, so the last row/column
+blocks are tails. With the opt-in tail-mask scheme (TL_ASCEND_TAIL_MASK)
+``T.tile.add`` is rewritten to a tail-aware helper that computes only over the
+valid rectangle; the unused UB gap is pad-filled and never written back.
+"""
+
+import tilelang
+import tilelang.language as T
+import torch
+
+tilelang.cache.clear_cache()
+
+pass_configs = {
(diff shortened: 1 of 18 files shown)


benyang0506 (approved):
lgtm

LLMZhangYC (commented):
approve

LLMZhangYC (approved):
approve

PR #1361, opened 2026-07-14
[Add] block_sparse_mqa_attn_expert_test_for_a5

handsomeRobotSK:
Summary
This PR adds two Block Sparse Multi-Query Attention (MQA) kernel implementations for the Ascend NPU, in Expert mode with full manual control of memory hierarchy, synchronization, and compute pipelines.
1. block_sparse_mqa_attn_expert_test.py — Base Expert Kernel (workspace path)
- Algorithm: For each query token, gather K blocks via top-k block indices, compute Q·K^T via MMA, apply ReLU + weight-mul + reduce-sum → logits, mask out-of-range positions with -inf.
- Memory path: C scope MMA → L0C → T.copy → GM workspace → V scope reads workspace → UB.
- Pipeline: Double-buffered L0B/L0C/K L1, 4-stage SW pipeline. Each kernel block handles a token pair (token_a / token_b) with paired-ping-pong L0A/L0B/L0C.
- Scheduling: Arbitrary grid_size (== cube core count), per-core num_pairs_bx guards tail cores.
- Sync: All flags and cross-scope handshakes manually managed (16 signal IDs in C scope, 3 in V scope).
- Includes: Golden reference (torch einsum) + test harness.
2. block_sparse_mqa_attn_expert_test_cv.py — A5 CV通路 Variant (no-workspace)
- Same algorithm as the base, but eliminates the GM workspace round-trip entirely.
- Key difference: Uses A5-exclusive T.copy_op.copy_cv_experiment() to DMA L0C tiles directly to UB via the FIX pipeline (TMOV). No workspace_1 argument.
- requires 4 separate per-block s_ub buffers (vs. single merged s_ub_4x in base) to avoid n_outer slot collisions.
- Adds a V→C buffer-free credit (CROSS_V2C) for back-pressure, preventing C from overwriting tiles V hasn't consumed yet.
- token_a tiles routed to vector core 0 (SingleVec0), token_b to core 1 (SingleVec1).
- A5-only: guarded by determine_platform() — gracefully exits on non-A5 platforms.
- All other logic (MMA pipeline, ReLU / row_expand_mul / reduce_sum / mask / output) identical to base.
Verification
Both kernels include inline golden validation (ref_block_sparse_mqa_attn) with assert_close(rtol=1e-2, atol=1e-2) and pass successfully.

### DIFF
--- a/examples/HISA/block_sparse_mqa_attn_expert_test_for_a5.py
+++ b/examples/HISA/block_sparse_mqa_attn_expert_test_for_a5.py
@@ -1,28 +1,52 @@
 """
-Block Sparse MQA Attention Kernel (Expert Mode)
+Block Sparse MQA Attention Kernel (Expert Mode, A5 CV通路 / no-workspace)
 
-Manual CV scope, manual sync, explicit L0A/L0B/MMA.
-Pure expert mode: all auto passes disabled.
+A5-only variant of block_sparse_mqa_attn_expert_test.py.
 
-Flag conventions:
-  - MTE1↔MTE2 for L1 buffer management
-  - M↔MTE1 for L0A/L0B management
-  - M↔FIX for L0C management
-  - V↔MTE2 for UB buffer management
-  - V↔MTE3 for output buffer management
+The base kernel routes each [H, kv] score tile through a GM workspace:
+    C:  MMA -> L0C -> T.copy(l0c, workspace_1[token, n_i])   (L0C -> GM)
+    V:  T.copy(workspace_1[token, n_i], s_ub column)         (GM  -> UB)
+
+This version uses the A5-exclusive CV通路 (L0C -> UB direct via TMOV,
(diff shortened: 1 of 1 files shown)


fuhouyu-hw (approved):
lgtm

examples/HISA/block_sparse_mqa_attn_expert_test.py
```diff
@@ -35,14 +35,19 @@ def block_sparse_mqa_attn_return_logits(
     block_N: int = 8,
     num_stages: int = 2,  # noqa: ARG001
     threads: int = 2,  # noqa: ARG001
-    num_pairs: int = 20,
+    grid_size: int = 24,
```
benyang0506, line 38: Ascend does not use the concept of a grid, which is GPUs. It can be described instead in terms of block_num, task_num, or a similar execution unit abstraction.

LLMZhangYC (approved):
approve

PR #1383, opened 2026-07-15
[Fix] Support rank-zero sources in Ascend T.copy

Aux-724:
## Summary

- add an Ascend copy override that lowers default whole-buffer rank-zero sources copied to UB through scalar load/store operations
- delegate every non-scalar path and option to the existing copy implementation, including the latest `tmp` argument
- add real-NPU, dual-codegen, and mismatched-shape regression coverage

## Motivation

`T.copy` rejected a scalar `T.Tensor(())` source when the destination was a one-element UB buffer because `[]` and `[1]` failed structural equality. Merely flattening the source allows lowering, but emits a one-element DMA configuration that is invalid on Ascend hardware. The scalar path therefore uses the source buffer's one-element flat view and emits a scalar store into UB.

## Validation

- full CMake build with Ascend enabled against the latest `ascendc_pto` base
- clean three-way merge with the latest base
- `7 passed` in `test_tilelang_ascend_language_scalar_copy.py` on the merged tree
  - AscendC real-NPU execution for float16 and float32
  - AscendC and PTO lowering/codegen for float16 and float32
  - `(2,) -> (1,)` mismatch remains rejected
- existing float16 GM-to-UB-to-GM AscendC regression passed
- original Task 68 rank-zero epilogue passed on real NPU at shape `(128, 16384)` with max absolute error `0.0`
- Ruff check/format and `git diff --check` passed

### DIFF
--- a/testing/python/language/test_tilelang_ascend_language_scalar_copy.py
+++ b/testing/python/language/test_tilelang_ascend_language_scalar_copy.py
@@ -1,0 +1,86 @@
+import pytest
+import torch
+
+import tilelang
+import tilelang.language as T
+import tvm
+
+
+PASS_CONFIGS = {
+    tilelang.PassConfigKey.TL_ASCEND_AUTO_CV_COMBINE: True,
+    tilelang.PassConfigKey.TL_ASCEND_AUTO_CV_SYNC: True,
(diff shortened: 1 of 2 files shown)


fuhouyu-hw:
check your CI result

PR #1389, opened 2026-07-16
[Feat] Make Bisheng compile flags kernel-scoped (#1386)

### LINKED ISSUE #1386: [Feature] Make Bisheng compiler options kernel-scoped
## Feature Description

#1346 introduced `AscendSyncInsertVS` and paired it with disabling Bisheng auto-sync. The current implementation applies that policy by writing the following values into `os.environ` when VS sync is enabled:

```text
TL_CCE_AUTO_SYNC=off
TL_CCE_OPT_LEVEL=3
```

These variables are process-wide and are never restored. As a result, compiling one kernel changes how later, unrelated kernels are compiled.

Relevant code:

- [`process_default_pass_config()` mutates `os.environ`](https://github.com/tile-ai/tilelang-ascend/blob/6c4a0b1627b663bda0d67967b8f01de15bfcd201/tilelang/transform/pass_config.py#L164-L175)
- [`LibraryGenerator` reads the process-wide values](https://github.com/tile-ai/tilelang-ascend/blob/6c4a0b1627b663bda0d67967b8f01de15bfcd201/tilelang/jit/adapter/libgen.py#L108-L153)

## Use Case

The leak can be triggered by compiling two kernels in the same process:

1. Kernel A enables `TL_ASCEND_AUTO_SYNC_VS`; this sets `TL_CCE_AUTO_SYNC=off` and `TL_CCE_OPT_LEVEL=3`.
2. Kernel B disables `TL_ASCEND_AUTO_SYNC_VS`; nothing restores the environment.
3. Kernel B is still compiled with `--cce-auto-sync=off -O3`, although it did not request either option.

This makes compiler behavior depend on kernel compilation order. It also prevents a kernel from disabling CCE auto-sync without enabling `AscendSyncInsertVS`, which is required by expert kernels that manage synchronization themselves.

## Proposed Solution

Make Bisheng options kernel-scoped and pass them through the JIT pipeline, for example:

```python
@tilelang.jit(
    target="ascendc",
    compiler_options={
        "cce_auto_sync": False,
        "opt_level": 3,
    },
    pass_configs={
        tilelang.PassConfigKey.TL_ASCEND_AUTO_SYNC: False,
        tilelang.PassConfigKey.TL_ASCEND_AUTO_SYNC_VS: False,
    },
)
```

Expected behavior:

- Preserve #1346's default pairing: enabling VS sync may derive `cce_auto_sync=False` and `opt_level=3` for that kernel only.
- Explicit per-kernel options take precedence; existing environment variables remain compatibility fallbacks.
- Do not modify `os.environ` while processing pass configs.
- Include normalized compiler options in the kernel cache key.
- AscendC maps `cce_auto_sync=False` to `--cce-auto-sync=off`; PTO keeps its existing auto-sync behavior.

Acceptance case: after compiling kernel A with VS sync enabled, kernel B with VS sync disabled must use its own/default Bisheng options rather than A's options.

## Feature Type

- [x] Compiler optimization
- [x] Developer tooling

## Additional Context

Related: #1346

platelett:
Fixes #1386.

## How to review

- [Commit 1 — skip](https://github.com/tile-ai/tilelang-ascend/pull/1389/commits/87f02d8dda4ff567a9572e66dc67d175bca72db1): ruff 0.15.21 formatting + typing modernization. No behavior change.
- [Commit 2 — review this](https://github.com/tile-ai/tilelang-ascend/pull/1389/commits/57da5d1d99d1563ebd8e1c1a7bd624894b19a1d2): the kernel-scoped compile-flags implementation.
- [Commit 3 — reviewer follow-up](https://github.com/tile-ai/tilelang-ascend/pull/1389/commits/30dd9c307941e0e68b30f46367e04e0554f1185c): finalizes synchronization coverage in two files. The example contains only successful, automatically synchronized usage; the NPU-gated pytest owns the pass-on/pass-off negative regression.

## Problem

#1346 writes `TL_CCE_AUTO_SYNC` and `TL_CCE_OPT_LEVEL` into `os.environ` at pass-config time. These variables are process-wide and never restored, so compiling one kernel with VS sync ON leaks `-O3 --cce-auto-sync=off` into every subsequent compile in the same process. `TL_PTO_DEBUG` has the same scope problem.

## Solution

Derive Bisheng flags per kernel instead of mutating the environment. The public API is `compile_flags`, matching upstream [tile-ai/tilelang](https://github.com/tile-ai/tilelang/blob/main/tilelang/jit/__init__.py):

```python
@tilelang.jit(target="ascendc", compile_flags=["--cce-auto-sync=off", "-O3"])
@tilelang.jit(target="pto", compile_flags=["-D_DEBUG", "--cce-enable-print"])
```

### How flags are built

`resolve_compile_flags(target, pass_configs, compile_flags)` produces the final flag list:

1. **Framework-derived defaults** (`normalize_compiler_options`): `-O2` by default; the #1346 VS-sync pairing (`TL_ASCEND_AUTO_SYNC_VS` → `-O3 --cce-auto-sync=off`); then the `TL_CCE_AUTO_SYNC` / `TL_CCE_OPT_LEVEL` / `TL_PTO_DEBUG` env vars as lenient read-only fallbacks.
2. **Caller's `compile_flags`** appended last — Bisheng is last-wins for repeated flags, so caller flags override the defaults.

The resolved list is computed once in `compile()` and threaded through cache → `JITKernel` → `LibraryGenerator`, folded into the cache key. `os.environ` is never written.

### File-by-file guide

| File | What to look for |
|------|------------------|
| `transform/pass_config.py` | `normalize_compiler_options()` derives `{cce_auto_sync, opt_level, pto_debug}` from `pass_configs` + env. Removes `_process_for_auto_sync_insert_vs`. |
| `jit/adapter/libgen.py` | `resolve_compile_flags()` turns the dict into flags + appends caller flags. `LibraryGenerator` appends (whitespace-split, de-duplicated) instead of hardcoding. |
| `jit/__init__.py` | `compile()` calls `resolve_compile_flags` once; `jit()` / `_JitImplementation` thread `compile_flags`. |
| `jit/kernel.py` | `JITKernel` stores `compile_flags` and passes to adapter. |
| `cache/__init__.py`, `cache/kernel_cache.py` | `compile_flags` in the cache key. |
| `jit/adapter/cython/adapter.py` | Forwards `compile_flags` to `LibraryGenerator`. |
| `autotuner/param.py` | `CompileArgs` gains `compile_flags` + `platform`; `_load_kernel_from_disk` fixes the stale `from_database()` call (see below); `auto_gm_idx` now persisted. |
| `autotuner/tuner.py` | `set_compile_args` exposes `compile_flags` + `platform`; `@autotune` inherits both from `@tilelang.jit`. |

### PTO compatibility

- `--cce-auto-sync` is AscendC-only and never emitted for PTO (whose default already matches `=off`).
- `-D_DEBUG --cce-enable-print` is PTO-only and never emitted for AscendC.
- Env fallbacks are lenient: only `"off"` disables auto-sync, only `"1"` enables PTO debug. A malformed value never raises — matching the pre-existing behavior, so a stray env var cannot break a PTO compile.
- Caller `compile_flags` are passed through to Bisheng without target filtering. This is intentional: the caller explicitly chose them, and Bisheng silently tolerates cross-target flags.

### Auto-tuner disk-restore bugfix

`AutotuneResult._load_kernel_from_disk` called `JITKernel.from_database()` with a stale argument list — missing `platform`, `workspace_idx`, `auto_gm_idx` — so restoring a tuned kernel from disk raised `TypeError`. This PR supplies the missing args, persists/restores `auto_gm_idx`, and resolves `platform` via `determine_platform` (sim-mode restores load the correct simulator lib).

### Reviewer notes

- **Cache-key change**: the resolved flag list is now part of the key. Two kernels differing only in Bisheng flags will no longer alias to the same cached artifact (a one-time invalidation for existing Ascend kernel caches).
- **Not ported from upstream**: `func.attrs["tilelang_compile_flags"]` and `pass_configs[TL_DEVICE_COMPILE_FLAGS]` — neither mechanism exists in this repo; orthogonal to #1386.

### Reviewer follow-up: synchronization coverage

The final layout follows [fengz72's review guidance](https://github.com/tile-ai/tilelang-ascend/pull/1389#discussion_r3602086298):

- `examples/compile_flags/compile_flags_example.py` contains only successful usage. It has no `T.barrier_all()` or explicit execution scope, enables `TL_ASCEND_AUTO_SYNC`, and verifies both explicit and default Bisheng flags against PyTorch.
- `testing/python/language/test_ascend_compile_flags.py` owns the negative regression. The same dependency-chain kernel is compiled under identical Bisheng flags with `TL_ASCEND_AUTO_SYNC` enabled and disabled; the synchronized output matches PyTorch, while the intentionally unsynchronized output does not.

## Tests

- `testing/python/language/test_ascend_compile_flags.py` — 17 hermetic flag-resolution/JIT/cache/autotuner instances plus one NPU-gated synchronization regression.
- `examples/compile_flags/compile_flags_example.py` — positive-only on-device demonstration. Two synchronized kernels use explicit and default Bisheng flags respectively; both match PyTorch, verifying correctness and kernel-scoped flag isolation. Auto-collected by `bench_test.sh`.

### DIFF
--- a/examples/compile_flags/compile_flags_example.py
+++ b/examples/compile_flags/compile_flags_example.py
@@ -1,0 +1,83 @@
+"""Kernel-scoped Bisheng compile flags (issue #1386).
+
+Demonstrates the ``compile_flags`` argument of ``@tilelang.jit`` and shows that
+flags are resolved per kernel: passing ``--cce-auto-sync=off`` / ``-O3`` to one
+kernel does not change how a later kernel with default flags is compiled.
+
+Historically these were driven by the process-wide ``TL_CCE_AUTO_SYNC`` /
+``TL_CCE_OPT_LEVEL`` environment variables, which leaked across kernels compiled
+in the same process. ``compile_flags`` makes them kernel-scoped.
+
+Run: python compile_flags_example.py
+"""
+
+import argparse
+
+import torch
+
+import tilelang
+import tilelang.language as T
+
+tilelang.cache.clear_cache()
+
+parser = argparse.ArgumentParser(description="compile_flags example")
+parser.add_argument("--m", type=int, default=256, help="Matrix M dimension")
+parser.add_argument("--n", type=int, default=256, help="Matrix N dimension")
+args = parser.parse_args()
+
+M, N = args.m, args.n
+VEC_NUM = 2
+
+
+def vec_add(M, N, block_M, block_N, compile_flags, dtype="float"):
+    m_num = M // block_M
+    n_num = N // block_N
+
+    @tilelang.jit(out_idx=[-1], compile_flags=compile_flags)
+    def build():
+        @T.prim_func
+        def main(
+            A: T.Tensor((M, N), dtype),
+            B: T.Tensor((M, N), dtype),
+            C: T.Tensor((M, N), dtype),
+        ):
+            with T.Kernel(m_num * n_num, is_npu=True) as (cid, vid):
+                bx = cid // n_num
+                by = cid % n_num
+                a_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
+                b_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
+                c_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
+                # Explicit barriers -> correct regardless of --cce-auto-sync.
+                with T.Scope("V"):
+                    T.copy(A[bx * block_M + vid * block_M // VEC_NUM, by * block_N], a_ub)
+                    T.copy(B[bx * block_M + vid * block_M // VEC_NUM, by * block_N], b_ub)
+                    T.barrier_all()
+                    T.tile.add(c_ub, a_ub, b_ub)
+                    T.barrier_all()
+                    T.copy(c_ub, C[bx * block_M + vid * block_M // VEC_NUM, by * block_N])
+
+        return main
+
+    return build()
+
+
+torch.manual_seed(0)
+a = torch.randn(M, N).npu()
+b = torch.randn(M, N).npu()
+ref = a + b
+
+# Kernel A: per-kernel flags -> bisheng gets --cce-auto-sync=off and -O3.
+kernel_a = vec_add(M, N, 128, 256, compile_flags=["--cce-auto-sync=off", "-O3"])
+# Kernel B: default flags (no compile_flags) -> bisheng gets -O2, auto-sync on.
+# If Kernel A's flags leaked, Kernel B would be miscompiled.
+kernel_b = vec_add(M, N, 128, 256, compile_flags=None)
+
+print("init successful!")
+
+out_a = kernel_a(a, b)
+out_b = kernel_b(a, b)
+
+torch.testing.assert_close(out_a, ref, rtol=1e-2, atol=1e-2)
+torch.testing.assert_close(out_b, ref, rtol=1e-2, atol=1e-2)
+
+print("Kernel Output Match!")
--- a/examples/compile_flags/compile_flags_example.py
+++ b/examples/compile_flags/compile_flags_example.py
@@ -1,12 +1,21 @@
 """Kernel-scoped Bisheng compile flags (issue #1386).
 
-Demonstrates the ``compile_flags`` argument of ``@tilelang.jit`` and shows that
-flags are resolved per kernel: passing ``--cce-auto-sync=off`` / ``-O3`` to one
-kernel does not change how a later kernel with default flags is compiled.
+This example demonstrates two synchronization controls that serve different
+purposes:
 
-Historically these were driven by the process-wide ``TL_CCE_AUTO_SYNC`` /
-``TL_CCE_OPT_LEVEL`` environment variables, which leaked across kernels compiled
-in the same process. ``compile_flags`` makes them kernel-scoped.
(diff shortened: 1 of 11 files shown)


examples/compile_flags/compile_flags_example.py
```diff
@@ -0,0 +1,83 @@
...
+                bx = cid // n_num
+                by = cid % n_num
+                a_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
+                b_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
+                c_ub = T.alloc_ub((block_M // VEC_NUM, block_N), dtype)
+                # Explicit barriers -> correct regardless of --cce-auto-sync.
+                with T.Scope("V"):
+                    T.copy(A[bx * block_M + vid * block_M // VEC_NUM, by * block_N], a_ub)
+                    T.copy(B[bx * block_M + vid * block_M // VEC_NUM, by * block_N], b_ub)
+                    T.barrier_all()
```
fengz72: examples里面尽量别用T.barrier_all(), 可以开启自动插同步的pass

fengz72 (approved):

LLMZhangYC (approved):
lgtm

benyang0506 (approved):
lgtm

PR #1421, opened 2026-07-21
[Ascend] Support guarded axis-0 tail reductions in AscendC/PTO

cookiebai:
## Summary

This PR extends the tail-mask framework merged by #1360 with a guarded AscendC axis-0 reduction path.

It intentionally supports a strict, correctness-first subset rather than general tail reduction:

* backend: AscendC and `auto` lowering only;
* operations: `reduce_sum`, `reduce_max`, and `reduce_min`;
* input/output dtype: matching `float32`;
* rank/layout: clean 2D UB tiles whose encoded reduce shape matches the physical tile;
* axes: `0` and its equivalent negative-axis spelling `-2`;
* accumulation: `clear=true`;
* output layout: `[1, physical_col]`, with valid region `[1, valid_col]`.

PTO, last-axis reductions (`1`/`-1`), lower-precision reductions, accumulating reductions, explicit shape mismatches, and other unsupported or unproven contracts are not rewritten and remain on their native backend paths. Frontend-invalid output layouts continue to be rejected by the existing reduce shape validation.

## Relationship to #1360

#1360 introduced `TL_ASCEND_TAIL_MASK`, `AscendTailMaskPropagation`, tail-aware unary/binary/scalar helpers, and the hybrid `pad_value` fallback. It deliberately did not rewrite reductions.

This PR is based on the tail-mask framework merged by #1360 and contains only the reduction extension. It reuses the valid-region metadata produced by the existing GM-to-UB copy handling and does not duplicate the base tail-mask implementation.

## Implementation

For a tracked input region `[valid_row, valid_col]` with physical layout `[physical_row, physical_col]`, an eligible axis-0 reduction is rewritten to `tl.ascend_tail_reduce`.

The AscendC helper initializes the output from the first valid row, then applies vector `Add`, `Max`, or `Min` over each remaining valid row. Every operation processes only `valid_col` elements and uses `row * physical_col` as its source offset, so invalid rows and padded columns are never read.

The resulting tail state is:

```text
[valid_row, valid_col] -> [1, valid_col]
[physical_row, physical_col] -> [1, physical_col]
```

This preserves the column tail for downstream unary, binary, and scalar rewriting. Unsupported reductions clear the propagated output state and retain their native implementation.

The internal helper keeps `tmp`, `dim`, and `clear` in its ABI for compatibility with the existing reduction call. The validated implementation does not consume `tmp`; emitted calls are normalized to `dim=0` and are guarded to `clear=true`.

The unvalidated last-axis scalar helper path has been removed. Last-axis reductions remain on the existing native reduce path.

## Tests

Code-generation coverage verifies:

* `sum`, `max`, and `min` helper emission for axis `0` and frontend alias `-2`;
* axis-0 reduce-to-unary tail propagation;
* PTO backend fallback using an otherwise supported axis-0 contract;
* `TL_ASCEND_TAIL_MASK=false` fallback using an otherwise supported contract;
* orthogonal guards for `clear=false`, `float16`, and mismatched `real_shape`;
* last-axis native fallback and downstream tail-state clearing;
* presence of the native reduce marker in negative cases, rather than checking only for the absence of the tail helper.

NPU coverage includes `sum`, `max`, and `min`; axes `0` and `-2`; row-only tails; column-only tails; simultaneous row and column tails; one-valid-row and one-valid-column tails; tensors smaller than a tile; exact tiles; a 64-row physical tile; and NaN, infinity, and signed-zero inputs.

Pure reduction tests use a minimal Vector configuration and do not enable automatic CV combine or CV sync.

## Validation

* C++ sources compile through the TileLang object target.
* `ruff format --check` and `ruff check` pass for the modified Python tests.
* Python byte-compilation and `git diff --check` pass.
* Full NPU numerical execution was not run locally and will be validated by the NPU CI jobs.

### DIFF
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -2285,7 +2285,7 @@ void CodeGenTileLangAscend::TailReduceOpCodegen(const CallNode *op) {
   std::string dtype = getType(GetAccessPtrDtype(op->args[1].as<CallNode>()));
   std::string out = PrintBufferOffset(op->args[1].as<CallNode>());
   std::string src = PrintBufferOffset(op->args[2].as<CallNode>());
-  std::string tmp = PrintBufferOffset(op->args[3].as<CallNode>(), false);
+  std::string tmp = PrintBufferOffset(op->args[3].as<CallNode>());
   // Bind valid-region expressions first (see TailUnaryOpCodegen).
   std::string dim_str = PrintExpr(op->args[4]);
   std::string vrow = PrintExpr(op->args[5]);
--- a/src/target/codegen_ascend_pto.cc
+++ b/src/target/codegen_ascend_pto.cc
@@ -875,14 +875,16 @@ void CodeGenTileLangAscendPto::VisitExpr_(const CallNode *op,
     BinaryVecOpsCodegen(op, "TMINS");
 
     // --- tail-aware vector ops (AscendTailMaskPropagation) ---
-    // Only unary / binary / scalar are rewritten; reduce / broadcast / compare
-    // / select stay on the full-tile path (gap filled with pad_value).
+    // Broadcast / compare / select stay on the full-tile path (gap filled with
+    // pad_value).
   } else if (op->op.same_as(tl::ascend_tail_unary())) {
     TailUnaryOpCodegen(op);
   } else if (op->op.same_as(tl::ascend_tail_binary())) {
     TailBinaryOpCodegen(op);
   } else if (op->op.same_as(tl::ascend_tail_scalar())) {
     TailScalarOpCodegen(op);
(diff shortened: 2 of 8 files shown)


fuhouyu-hw (approved):
lgtm

benyang0506 (approved):
approve

ChaoyangJi:
这个是修复的什么问题? 是只修改了asendc后端还是pto后端你一起修复？为啥标题里写的是ascendc后端，但里面的修改有pto的修改？

cookiebai:
> 这个是修复的什么问题? 是只修改了asendc后端还是pto后端你一起修复？为啥标题里写的是ascendc后端，但里面的修改有pto的修改？

目前ascend c和pto都修复了，一开始描述里面只完成了ascendc，后面补了一版pto

PR #1424, opened 2026-07-21
【Feat】Add examples_experiment path and update CI

hedi515:
## What

Relocate 11 operators from `examples/` into a new `examples_experiment/`
directory, and extend the benchmark CI so those operators are still
exercised — both in full runs and in incremental PR runs.

## Why

The 11 operators were moved to `examples_experiment/` to separate
experimental kernels from the stable example set. Before this change the
CI (`examples/bench_test.sh` + `.github/workflows/ci_cd.yml`) only scanned
`examples/`, so the relocated operators were no longer tested by CI at all.

### `examples/bench_test.sh`
- Add `--experiment-dirs` and a second scan root (`../examples_experiment`).
- **Full mode** now scans both `examples/` and `examples_experiment/` into a
  single parallel pool and one unified pass/fail summary (result parsing in
  the workflow is unchanged).
- **Incremental mode** collects experiment dirs independently from example dirs.
- `--dirs` / `--experiment-dirs` now accept multiple space-separated dirs.
  Previously, because the workflow passes the value unquoted
  (`./bench_test.sh $TEST_DIRS_ARG`), only the *first* directory was kept and
  the rest were silently dropped — so a PR touching several operator dirs only
  ran the first one. Both flags now consume all following non-flag tokens.
- Guard `EXTRA_TASKS` against missing directories.

### `.github/workflows/ci_cd.yml`
- Detect `examples_experiment/<dir>/` changes as incremental experiment dirs
  (new `experiment_dirs` output).
- Pass `--experiment-dirs` through to `bench_test.sh`.
- Treat a run as full only when **both** `test_dirs` and `experiment_dirs` are
  empty (an experiment-only PR is no longer misclassified as a full run).
- Allow experiment-only PRs to skip the pytest suite (they can't affect
  `testing/python/`), consistent with how `examples/`-only PRs are handled.

## Behavior after this PR

| Trigger | Behavior |
|---|---|
| schedule / manual (full) | runs all `examples/` + `examples_experiment/` operators |
| PR touching `examples/<dir>/` | incremental: that example dir |
| PR touching `examples_experiment/<dir>/` | incremental: that experiment dir |
| PR touching both | incremental: both dirs |
| PR touching core files (`src/`, `tilelang/`, …) | full (incl. experiment) |
| experiment-only doc change | skipped |

## Testing

Validated locally (collection logic only; operators need an Ascend NPU to run):
- `bench_test.sh`: full (both roots), single/multi example dirs, single/multi
  experiment dirs, mixed CI-style invocation, quoted-single-arg compatibility,
  and error handling for `--dirs` with no value — all collect the expected
  script sets.
- Change-detection classification for all 9 scenarios in the table above.


fuhouyu-hw (approved):
lgtm

benyang0506 (approved):
approve

PR #1445, opened 2026-07-23
[Fix] Handle rank-changing T.Parallel copies on Ascend

Aux-724:
## Problem

`AscendLowerParallelToVector` accepts a one-dimensional `T.Parallel` destination as vectorizable even when the direct `BufferLoad` source has a different rank. For a 2D UB-to-1D UB row projection such as:

```python
for lane in T.Parallel(lanes):
    dst[lane] = src[rows - 1, lane]
```

`GenerateAscendCopy` takes its 2D-source path, then unconditionally reads `dst_gm->shape[1]`. Since the destination rank is one, compilation fails with:

```text
InternalError: IndexError: indexing 1 on an array of size 1
```

Changing the extent to a source dimension would not be correct: row and column projections require different region geometry, and the current vector-copy helper no longer retains enough information to distinguish them.

## Fix

Reject rank-changing direct copies in vector-copy decomposition and reuse the existing correctness-first `T.Parallel` to serial fallback introduced for #1194. Matching-rank vector copies and the existing 1D-to-2D broadcast lowering remain unchanged.

Add an NPU regression that projects the final row of a `(256, 8)` float32 UB into a rank-1 UB and checks the result against the source row.

## Validation

- Before the fix, the new regression reproduces the exact `AscendLowerParallelToVector -> GenerateAscendCopy -> Array::operator[]` failure.
- `cmake --build build -j64`
- `pytest testing/python/language/parallel/test_tilelang_ascend_language_parallel_ub_stride.py -q -s`: **6 passed**
- Four existing row/column scalar-buffer broadcast tests: **4 passed**
- Executor-compatible `msprof op` run on Ascend NPU:
  - max absolute error: `0.0`
  - end-to-end wrapper latency: `1.1905607767403126 ms`
  - captured kernel task duration: `11.900238 us`
  - matched op: `row_projection_profile_kernel_mix_aic`

### DIFF
--- a/src/transform/ascend_lower_parallel_to_vector.cc
+++ b/src/transform/ascend_lower_parallel_to_vector.cc
@@ -901,8 +901,9 @@ private:
     if (auto load = expr.as<BufferLoadNode>()) {
       Buffer input_buffer = load->buffer;
 
-      // Rank-changing copies require projection-aware regions, which the
-      // vector copy helper cannot represent. Use the scalar serial fallback.
+      // GenerateAscendCopy indexes source and destination shapes symmetrically.
+      // Rank-changing copies require projection-aware regions, which the vector
+      // copy helper cannot represent. Use the scalar serial fallback.
       if (input_buffer->shape.size() != output_buffer->shape.size()) {
         return false;
       }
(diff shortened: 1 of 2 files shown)


Aux-724:

PR #1451, opened 2026-07-23
[Fix] Substitute affine T.Parallel outer indices on Ascend

Aux-724:
## Summary

- retain the known outer T.Parallel axis when a BufferStore index is affine, such as base + g
- let the existing broadcast/serial fallback replace that axis instead of emitting a free TIR variable
- cover both 2D and 3D store-plan detection
- add an Ascend NPU regression nested under a serial loop

## Root cause

DetectVectorPlan tried to recover the outer variable with index.as<VarNode>(). That returns null for affine expressions such as base + g. If expression analysis later disables 2D vectorization, the fallback creates outer_broadcast_idx but has no variable to replace, so generated code still references the unbound g and codegen fails with Find undefined Variable g.

## Validation

- pre-fix minimal regression: reproducible Find undefined Variable g
- full TileLang C++ build with Ascend targets
- pytest test_tilelang_ascend_language_parallel_mixed.py::test_parallel_affine_outer_index (NPU: 1 passed)
- full parallel mixed test file (NPU: 5 passed)
- ruff format/check; clang-format reports no content changes
- three-way merge-tree check against current origin/ascendc_pto: no conflicts

### DIFF
--- a/src/transform/ascend_lower_parallel_to_vector.cc
+++ b/src/transform/ascend_lower_parallel_to_vector.cc
@@ -488,14 +488,15 @@ private:
       plan->inner_vec_len = inner_vec_len;
       plan->outer_extent = element_count / inner_vec_len;
 
-      // Try to extract outer variable from the outer index
-      plan->outer_index_var = store->indices[0].as<VarNode>();
-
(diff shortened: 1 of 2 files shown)


Aux-724:

PR #1452, opened 2026-07-23
[Fix] Flatten high-rank and compact Ascend GM/UB copies

Aux-724:
## What this fixes

Ascend GM<->UB lowering uses one 2D DMA, and high-rank/ping-pong UB regions must describe the selected region rather than the whole physical allocation.

## Changes

- flatten statically contiguous high-rank leading dimensions into the DMA row count
- preserve constant GM strides for trailing singleton dimensions
- use region rows for the copy helper template, avoiding writes across adjacent ping-pong slots
- support rank-1 GM slices copied to/from higher-rank UB tiles by deriving rows/columns from the UB region
- support compact full-row non-32-byte UB copies with one contiguous DataCopyPad burst
- keep bounds, contiguity, and non-compact unaligned-layout checks
- add NPU regressions for contiguous high-rank copies, singleton strides, and compact unaligned rows

## Validation

- C++ build: `cmake --build build -j4`
- high-rank copy regression: 5 passed on NPU
- `examples/elementwise/elementwise_add_pipeline.py`: Kernel Output Match
- `examples/causal_conv1d/causal_conv1d.py`: Batch Kernel Output Match
- GQA correctness and perf scripts: passed
- `examples/deepseek_v4/hc_split_sinkhorn.py`: Kernel Output Match
- xattention and xattention_paged kernel generation/execution: passed locally


Aux-724:

Aux-724:
Follow-up fix pushed in 15a3db2b. The failed alloc_codegen cases exposed a real regression: dynamic row extents in a rank>2 GM buffer (for example K_cache[1, prefix_len, 64]) were forced through the static high-rank flattening path. The lowering now keeps static high-rank copies on the strict validated path, while dynamic high-rank GM<->rank-2-UB copies use the existing runtime row-mask path; dynamic high-rank operands on both sides still fail explicitly. Local build and test_tilelang_ascend_language_alloc_codegen.py pass (2/2). /re-test

Aux-724:
Follow-up fix pushed in 15a3db2b. The failed alloc_codegen cases exposed a real regression: dynamic row extents in a rank>2 GM buffer (for example K_cache[1, prefix_len, 64]) were forced through the static high-rank flattening path. The lowering now keeps static high-rank copies on the strict validated path, while dynamic high-rank GM<->rank-2-UB copies use the existing runtime row-mask path; dynamic high-rank operands on both sides still fail explicitly. Local build and test_tilelang_ascend_language_alloc_codegen.py pass (2/2). /re-test

Aux-724:
Follow-up fix pushed in 2bd25b2e. The NPU benchmark showed that valid high-rank UB tiles may use a non-zero first-dimension offset for double buffering (for example `(mm + 1) % 2`), while the previous check incorrectly required every UB range minimum to be zero. The check now accepts offset tiles when `min + extent <= buffer.shape`, while dense interior dimensions remain validated by the flattening logic. Local `elementwise_add_pipeline.py` and `causal_conv1d.py` both pass; high-rank lowering regressions pass. `/re-test` has been requested.

PR #1453, opened 2026-07-23
[Fix] Handle compact unaligned multi-row AscendC UB copies

Aux-724:
## What this fixes

AscendC `DataCopyPad` expresses the local UB row stride in 32-byte blocks. A compact multi-row UB such as `float16[2, 9]` has an 18-byte logical row, so the regular 2D burst can silently corrupt later rows even when the source and destination are both contiguous.

## Changes

- detect compact full-row GM-to-UB and UB-to-GM copies whose logical row is not 32-byte aligned
- issue one contiguous DMA burst for those copies, avoiding unaligned per-row addresses
- preserve the existing 2D path for strided or tail slices used by the vector kernels
- replace the compile-failure regression with an NPU load -> UB add -> store correctness test

This only changes AscendC GM/UB copy lowering; PTO behavior is unchanged.

## Validation

- `python3 -m pytest -q testing/python/language/test_tilelang_ascend_language_elementwise.py::test_unaligned_multirow_full_rows_supported`
- `test_vec_select`, `test_vec_select_scalar`, and AscendC tail-mask regression passed on NPU
- PTO tail-mask test remains blocked by the local PTO compiler missing `__biasbuf__`/`aicore` definitions

### DIFF
--- a/src/tl_templates/ascend/common.h
+++ b/src/tl_templates/ascend/common.h
@@ -217,6 +217,8 @@ copy_gm_to_ub(LocalTensor<T> dstTensor, GlobalTensor<T> srcTensor,
       WaitFlag<HardEvent::V_MTE2>(0);
     }
   }
+  static_assert(dstM <= 1 || (dstN * sizeof(T)) % 32 == 0,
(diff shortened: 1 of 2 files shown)


Aux-724:

PR #1457, opened 2026-07-24
[Example] Ascend: sparse flash MLA (DeepSeek) with shared paged KV

yangqiang2018:
## What

Add an Ascend example under `examples/deepseek_v4/sparse_flash_mla/` implementing sparse flash MLA with a shared paged KV cache, in three variants:

- `swa` — sliding window over the original KV
- `hca` — original + compressed KV
- `csa` — original + compressed KV + top-k sparse indices

Files:

- `sparse_flash_mla_kernel.py` — TileLang kernels (three `@tilelang.jit` builders)
- `sparse_flash_mla_api.py` — the `sparse_flash_mla(...)` entry point
- `sparse_flash_mla_golden.py` — reference implementation + accuracy checks

## Validation

`python sparse_flash_mla_api.py` runs the three variants against the golden reference and prints `Kernel Output Match!`.

### DIFF
--- a/examples/deepseek_v4/sparse_flash_mla/sparse_flash_mla_api.py
+++ b/examples/deepseek_v4/sparse_flash_mla/sparse_flash_mla_api.py
@@ -1,0 +1,272 @@
+# Copyright (c) Tile-AI Corporation.
+# Licensed under the MIT License.
+
+from __future__ import annotations
+
+from typing import Optional, Tuple, Union
+
+import torch
+
+from sparse_flash_mla_kernel import build_sparse_flash_mla, DEFAULT_BLOCK_I, DEFAULT_CORE_NUM
+
+
+_KERNEL_CACHE: dict = {}
+
+
+def _torch_to_tilelang_dtype(t: torch.dtype) -> str:
+    if t == torch.bfloat16:
+        return "bfloat16"
+    if t == torch.float16:
+        return "float16"
+    raise ValueError(f"unsupported torch dtype {t}")
+
+
+def _get_kernel(
+    *,
+    batch: int,
+    max_seq: int,
+    total_tokens: int,
+    ori_block_num: int,
+    ori_block_size: int,
+    ori_table_len: int,
+    cmp_block_num: int,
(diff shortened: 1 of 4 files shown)


examples/deepseek_v4/sparse_flash_mla/sparse_flash_mla_api.py
```diff
@@ -0,0 +1,272 @@
+# Copyright (c) Tile-AI Corporation.
+# Licensed under the MIT License.
```
LLMZhangYC, line 2: Please add bench_mark.md.

yangqiang2018: Added `bench_mark.md` in the example directory — input parameters, measurement method, the AscendC-vs-TileLang table for all six scenarios (swa / hca / csa × prefill / decode), and the optimization strategies the kernel uses.

Measured with `msprof` **device Task Duration** from `op_summary`, first launch dropped (cold start) and the rest averaged. Both sides run the same shapes and dtype: bf16, B=1, 64 query heads over 1 KV head, D=512, paged block size 128, sliding window 127; 8192-token prefill, and 1-step decode against an 8193-token KV cache.

| Scenario | AscendC | tileLang | Performance Ratio (AscendC/tileLang) |
|------|------|------|------|
| swa prefill | 1850.954u | 1490.930u | 124.1% |
| hca prefill | 3005.296u | 3141.853u | 95.7% |
| csa prefill | 6882.407u | 6022.700u | 114.3% |
| swa decode | 23.916u | 13.824u | 173.0% |
| hca decode | 24.007u | 18.644u | 128.8% |
| csa decode | 38.759u | 31.659u | 122.4% |

A ratio above 100% means the TileLang kernel is faster than the hand-written AscendC operator it is ported from. Five of the six scenarios are at or above parity; `hca prefill` is the one that still trails, at 95.7%.

ChaoyangJi (approved):
lgtm

LLMZhangYC (approved):
approve

PR #1477, opened 2026-07-27
[Test] Add flash_attn_bhsd Pytest migration sample

dlaerbinjwy:
Builds on #1459. Three independent commits — take or drop separately.

## 1. flash_attn_bhsd sample

Same pattern as batch_gemm: importlib load with `sys.argv` replaced, so
the example file is unchanged. Passes on hardware; tightening the
tolerance to 1e-5 fails at 2^-11 (the float16 floor), so the assertion
is load-bearing.

## 2. An example test can stop running without CI noticing

`-not -name "test_*.py"` drops example tests from the legacy runner, and
Pytest only collects manifest entries. A file in neither place runs
nowhere while CI stays green.

`examples/gqa_fwd_varlen/test_gqa_fwd_varlen.py` is in that state: it
passed in the pre-change CI log, the find filter now excludes it, it is
not in the manifest, and no shell script references it.

`check-orphans` fails when an example `test_*.py` is covered by neither
the manifest nor an example shell script. The shell exemption keeps
`gemm_aot` and `torch_tl_ascend` from being flagged — `bench_test.sh`
still collects their `.sh` entry points. Reports exactly one file today.

The manifest now registers `gqa_fwd_varlen`; its 4 tests pass in 69s.

## 3. Pytest worker count

`-n 1` replaced the `-n 4` / `-n 8` this file used before, while
`bench_test.sh` still uses `MAX_JOBS=8`.

Measured on an Ascend node: `testing/python/` at `-n 8` finishes in 884s
with 1253 passed / 0 failed; the three operator tests at `-n 8` also
pass. The node exposes only two NPUs, so device access already
serializes at the driver level. Roughly 5x faster than `-n 1`.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -57,9 +57,9 @@ jobs:
       merged: ${{ steps.pr_info.outputs.merged || steps.pr_status.outputs.merged || steps.default_changes.outputs.merged }}
       test_dirs: ${{ steps.detect_changes.outputs.test_dirs || steps.default_changes.outputs.test_dirs }}
       experiment_dirs: ${{ steps.detect_changes.outputs.experiment_dirs || steps.default_changes.outputs.experiment_dirs }}
-      pytest_files: ${{ steps.detect_changes.outputs.pytest_files || steps.default_changes.outputs.pytest_files }}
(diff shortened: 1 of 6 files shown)


LLMZhangYC (approved):
lgtm

PR #1493, opened 2026-07-29
[Test]: add non-intrusive pytest coverage for example kernels

aynewsd:
Add pytest wrappers for deepseek_v4, xllm_kernels, pos_embedding, and examples_experiment kernels. The tests load example modules via importlib with argv isolation, preserve original kernel/reference flows, and expose built-in case suites through pytest parametrization.

Covered examples include examples/deepseek_v4, examples/pos_embedding, examples/xllm_kernels,
 as well as examples_experiment/blocksparse_gemm, examples_experiment/dequantize_gemm, and examples_experiment/convolution.


aynewsd:
@dlaerbinjwy

PR #1512, opened 2026-07-31
[Feat]: Add fused‑moe example

wch0810:
## Description

Add DeepSeek-style FusedMoE (Fused Mixture of Experts) Forward kernel, implemented in `example_fusedmoe.py` based on TileLang (pure GEMM + PyTorch post-processing).
Add layered test suite with integrated performance benchmark (`do_bench` + `msprof op`).

Strategy: NPU kernels do GEMM only; Vector ops (SiLU, mul) offloaded to PyTorch host. GEMM outputs fp16 directly (T.copy auto-casts L0C fp32 → GM fp16). Uses `combineCV` + `auto_sync` passes (no manual `T.Scope`/`T.barrier_all`).

Tested on Ascend NPU, fp16, d_hidden=7168, d_expert=2048, n_routed_experts=1, n_experts_per_token=4, batch_size=1, seq_len=8192. Precision verified against PyTorch golden (atol=1e-2, rtol=1e-2).

## Golden Config (d_hidden=7168, d_expert=2048, n_routed=1, top_k=4, batch=1, seq=8192, fp16)
| Kernel                              | Seq | d_hidden | d_expert | Latency (ms) | TFLOPS | Max Diff |
| ----------------------------------- | --- | -------- | -------- | ------------ | ------ | -------- |
| TileLang Shared expert              | 8192 | 7168    | 2048     | 9.35         | 77.0   | 3.91e-03 |
| TileLang Routed expert              | 8192 | 7168    | 2048     | 11.95        | 60.3   | 3.91e-03 |
| TileLang Pipeline (shared + routed) | 8192 | 7168    | 2048     | 21.24        | 67.9   | 3.91e-03 |
| baseline                        | 8192 | 7168    | 2048     | 37.50        | —      | —        |

- TileLang Pipeline latency accounts for 0.57× of baseline, **43% faster than Baseline**.

## Notes
- Pure GEMM kernels (single/dual/grouped/grouped_dual) using T.alloc_L1/L0C + combineCV pass (auto sync).
- **T.mma + L0A double buffer** for dual_gemm: overlaps MTE1(load L0A[k+1]) with M(mma[k]), reducing per-tile time by 49%. The overall wall-time gain is partially offset by fixed-core grid-stride scheduling overhead.
- **Fixed Core + grid-stride loop**: by-major tile ordering for L2 cache reuse (Weight 3.5MB << 192MB L2). L2 hit rate: 86% → 99.4%.
- Dual GEMM fusion: gate+up GEMM fused (Input read once), halving GM bandwidth.
- GEMM direct fp16 output: T.copy auto-casts L0C fp32 → GM fp16.
- Block sizes: BLOCK_M=64, BLOCK_N=256, BLOCK_K=128.
- Vector ops offloaded to host (workaround for unstable Cube→Vector cross_flag sync).
- Limitations: BLOCK_M fixed at 64 (host_preprocess alignment); Vector ops not fused into kernel.

### DIFF
--- a/examples/fusedmoe/example_fusedmoe.py
+++ b/examples/fusedmoe/example_fusedmoe.py
@@ -1,0 +1,1298 @@
+"""FusedMoE: DeepSeek-style Fused Mixture of Experts for NPU (TileLang-Ascend).
+
+Strategy: pure GEMM kernels + PyTorch post-processing.
+  - Each kernel does GEMM only (proven to work on NPU)
+  - T.alloc_L1/L0C + combineCV pass (auto sync, no manual T.Scope/barrier_all)
+  - SiLU + element-wise mul done in PyTorch between kernel calls
+  - GEMM outputs fp16 directly (T.copy auto-casts L0C fp32 -> GM fp16)
+
+Contains:
+  - Generic single-GEMM and grouped-GEMM kernels
+  - Dual-GEMM kernels (gate+up fused, Input read once)
+  - Shared expert and routed expert pipelines
+  - Golden functions (PyTorch reference)
+  - Host preprocessing (gating + token routing)
+"""
+
+import torch
+import torch.nn.functional as F
+
+import tilelang
+import tilelang.language as T
+
+# Block sizes (all at hardware limits)
+BLOCK_M = 64
+BLOCK_N = 256
+BLOCK_K = 128
+
+# Developer mode: enable combineCV + auto sync passes
+pass_configs = {
+    tilelang.PassConfigKey.TL_ASCEND_AUTO_CV_COMBINE: True,
+    tilelang.PassConfigKey.TL_ASCEND_AUTO_CV_SYNC: True,
+    tilelang.PassConfigKey.TL_ASCEND_AUTO_SYNC: True,
+}
+
+
+# ============================================================================
+# Generic Single-GEMM Kernel (follows grouped_gemm example pattern exactly)
+# ============================================================================
+@tilelang.jit(out_idx=[-1], pass_configs=pass_configs)
+def single_gemm_kernel(M, N, K, dtype="float16"):
+    """Generic single GEMM: Output = Input @ Weight^T (fp32 accumulation).
+
+    Uses the proven grouped_gemm pattern:
+      - T.alloc_L1 for input/weight buffers
+      - T.alloc_L0C for accumulation
+      - combineCV pass (auto sync, no manual T.Scope/barrier_all)
+      - combineCV pass (auto sync)
+    """
+    block_M = BLOCK_M
+    block_N = BLOCK_N
+    block_K = BLOCK_K
+    accum_dtype = "float32"
+
+    m_num = (M + block_M - 1) // block_M
+    n_num = (N + block_N - 1) // block_N
+
+    @T.prim_func
+    def kernel(
+        Input: T.Tensor([M, K], dtype),  # type: ignore
+        Weight: T.Tensor([N, K], dtype),  # type: ignore
+        # M2: Output fp16 directly (T.copy auto-casts L0C fp32 → GM fp16)
+        Output: T.Tensor([M, N], dtype),  # type: ignore
+    ):
+        with T.Kernel(m_num * n_num, is_npu=True) as (cid, _):
+            bx = cid // n_num
+            by = cid % n_num
+
+            A_L1 = T.alloc_L1((block_M, block_K), dtype)
+            B_L1 = T.alloc_L1((block_N, block_K), dtype)
+            C_L0 = T.alloc_L0C((block_M, block_N), accum_dtype)
+
+            loop_k = T.ceildiv(K, block_K)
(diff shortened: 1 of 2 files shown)


examples/fusedmoe/example_fusedmoe.py
```diff
@@ -0,0 +1,661 @@
...
+    ):
+        with T.Kernel(m_num * n_num, is_npu=True) as (cid, _):
+            bx = cid // n_num
+            by = cid % n_num
+
+            A_L1 = T.alloc_L1((block_M, block_K), dtype)
+            B_L1 = T.alloc_L1((block_N, block_K), dtype)
+            C_L0 = T.alloc_L0C((block_M, block_N), accum_dtype)
+
+            with T.Scope("C"):
```
LLMZhangYC: Use combineCV pass in this file and remove all T.Scope.

examples/fusedmoe/example_fusedmoe.py
```diff
@@ -0,0 +1,661 @@
...
+                loop_k = T.ceildiv(K, block_K)
+                for k in T.serial(loop_k):
+                    T.copy(Input[bx * block_M, k * block_K], A_L1)
+                    T.copy(Weight[by * block_N, k * block_K], B_L1)
+                    T.barrier_all()
```
LLMZhangYC: Remove all barrier_all() and use auto_sync or specific set/wait flag if necesary.

examples/fusedmoe/test_fusedmoe.py
```diff
@@ -0,0 +1,1067 @@
...
+    wdr,
+    expert_weights,
+    block_metadata,
+):
+    """Benchmark shared/routed/pipeline kernels with do_bench.
+
+    Defined outside the bench loop so lambdas capture function parameters
+    (not loop variables), avoiding ruff B023.
+    """
+    shared_ms = do_bench(
```
benyang0506, line 938: Please benchmark it with `msprof op` instead of `do_bench`, since `msprof op` should provide more accurate kernel-level performance measurements.

benyang0506:
Could you please investigate why TFLOPS is only around 62 when the `seq_len` and `dim` are large, given that the A2/A3 theoretical compute capability is 364 TFLOPS? Please continue analyzing and optimizing the implementation.

wch0810:
> Could you please investigate why TFLOPS is only around 62 when the `seq_len` and `dim` are large, given that the A2/A3 theoretical compute capability is 364 TFLOPS? Please continue analyzing and optimizing the implementation.

I have completed profiling and optimization based on msprof.
The main bottleneck is that the kernel is memory-bound instead of compute-bound.

1. Root Cause
The theoretical 364 TFLOPS is full-chip peak Cube compute throughput.
Our MoE workload has frequent GM↔L2↔L0 data movement. MTE2 memory transfer originally occupied ~60% of pipeline cycles, and Cube units cannot be fully saturated.

2. Applied Optimizations
- Fixed-core grid-stride scheduling to reorder tile access, lifting L2 cache hit rate from 86.2% → 99.4%
- L0A double buffering with T.mma to overlap data loading and computation, eliminating scalar stalls
- Enabled combineCV + auto_sync passes to remove redundant synchronization overhead

3. Result
Per-tile execution latency reduced by 49%.
End-to-end pipeline TFLOPS improved from ~62 → 67.9, wall latency down to 21.24ms.
The optimized kernel still outperforms the baseline by 43%.

Further large jumps in utilization are limited by the inherent memory bandwidth pressure of this MoE pattern.
All detailed profiling metrics and test data are documented in the PR description.

examples/fusedmoe/test_fusedmoe.py
wch0810: Done. Added `--profiler msprof` argument to test_fusedmoe.py.
`do_bench` is kept as default for stable CI execution without extra CANN profiling dependency,
while msprof mode is available for hardware-level pipeline analysis including L2 hit rate, cube utilization and pipeline breakdown statistics.

examples/fusedmoe/example_fusedmoe.py
wch0810: Done. All `T.barrier_all()` calls have been deleted.
- Single/grouped gemm kernels rely on `auto_sync` for pipeline synchronization.
- Optimized dual_gemm kernel uses explicit `set_flag/wait_flag` for L0A double buffer pipelining as suggested.
Verified no remaining `T.barrier_all(` in latest code.

examples/fusedmoe/example_fusedmoe.py
wch0810: Done. All `T.Scope()` are removed in the codebase.
`combineCV` and `auto_sync` passes are enabled through pass_configs for automatic CV synchronization.
All four gemm kernels apply this configuration. Verified that there are zero `T.Scope(` occurrences in current code.

benyang0506 (approved):
lgtm

LLMZhangYC (approved):
approve

PR #1542, opened 2026-08-04
[Example] Group group_norm and cummin under examples/cannbench

hedi515:
Collects the CANN benchmark operators under a single examples/cannbench directory instead of leaving them at the top level of examples/, so the benchmark set can grow without scattering unrelated operators around it.

Pure relocation: both files move byte-for-byte, and nothing in the tree references the old paths, so no imports, docs or CI configuration need updating.

The two operators share the structure the benchmark set is built around — several kernels written per case type, selected by a host dispatch function from the input shape and dtype:

- group_norm: serial vs. cpipeline, chosen on s_num and the channel tiling
- cummin: a transposing v1 path vs. a transpose-free v2 path, chosen on whether the trailing dimension is wide enough to vectorise


ChaoyangJi (approved):
lgtm

LLMZhangYC (approved):
approve

PR #1702, opened 2026-08-27
feat(ascendc): add native A5 target support

zhshgmail:
## Summary

- add native AscendC A5 target selection (`dav-3510`, `CATLASS_ARCH=3510`) while retaining A2/A3 `dav-2201` / `CATLASS_ARCH=2201`
- update the bundled CATLASS gitlink to the stable `v1.5.0` object `83073fedcac3d9df5b83f701b2cc00ec6a94f984`
- adapt the Ascend templates to CATLASS v1.5 tensor/layout APIs, including A5 L0A layout and tail/origin-preserving tiled layouts
- keep source-only codegen tests device-independent while applying the same target defaults and PassContext as the production compile path
- update and guard the documented manual A2 Bisheng command

## Verification

- `TL_PLATFORM=A3 python -m pytest -p no:cacheprovider -q testing/python/language/test_ascend_compile_flags.py testing/python/language/test_tilelang_ascend_language_alloc_codegen.py`: **35 passed, 1 skipped**
- no-NPU GEMM AOT compile (`m=128, n=256, k=64`) for native AscendC platforms **A5, A2, and A3**: all `rc=0`
- `git diff --check`: passed
- independent focused review: P0/P1/P2 = 0/0/0

## Not run

- repository-wide full hook/test suite was not run
- no NPU runtime or device execution was performed

### DIFF
--- a/docs/notebook/aot_jit.ipynb
+++ b/docs/notebook/aot_jit.ipynb
@@ -204,6 +204,7 @@
     "%%bash\n",
     "# 手动编译 .so\n",
     "bisheng --npu-arch=dav-2201 \\\n",
+    "-DCATLASS_ARCH=2201 \\\n",
     "-O2 \\\n",
     "-std=c++17 \\\n",
     "-xasc \\\n",
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -9,10 +9,10 @@
 #include <tvm/arith/analyzer.h>
 #include <tvm/ir/transform.h>
 #include <tvm/runtime/registry.h>
+#include <tvm/tir/analysis.h>
+#include <tvm/tir/expr.h>
 #include <tvm/tir/index_map.h>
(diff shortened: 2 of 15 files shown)


zhshgmail:
Repair-forward update: head `2b5cc9b705e3142524126d4f2f727498896d596a` restores the preserved A2/A3 linear-UB behavior while keeping strict A5 usable-capacity enforcement. Added a causal `3 x 64 KiB` regression for both A2 and A3. Local verification: incremental build PASS; A2/A3/A5 boundary subset `7 passed`; compile/architecture suite `40 passed, 1 skipped`; `git diff --check` and Ruff format PASS. Fresh non-author review on the exact head/tree returned ACCEPT (`P0=0, P1=0, P2=0`). This closes the prior `ascend_910b elementwise_add.py` cause (third 64 KiB buffer at offset 131072 versus the historical 196352-byte guard-band limit); it does not claim real-A5 precision/performance closure.

PR #1706, opened 2026-08-28
fix(pto): preserve packed mask stride after UB reuse

cookiebai:
## What

- Recover the logical packed-mask shape when memory planning reuses a differently typed UB allocation.
- Preserve the required 32-byte PTO predicate row stride instead of inheriting the backing tensor shape.
- Add codegen regression coverage for both compare and select mask views.

## Why

For the mixed unary/scalar-select tail pipeline with `float16` and PTO, memory planning may reuse a `[4, 64]` half UB allocation for the short-lived `uint8` predicate mask.

The PTO codegen previously inherited the backing allocation's 64-column shape, while the packed predicate requires an 8-byte logical width aligned to a 32-byte UB row stride. This corrupted values in the final tail block and caused:

`test_mixed_unary_scalar_select_tail[float16-pto]`

to report 3 mismatched elements.

## Validation

- Incremental C++ build succeeded.
- `test_tilelang_ascend_language_tail_mask_codegen.py`: 127 passed.
- Generated PTO mask views are now `TileUbDataND<uint8_t, 4, 32, ...>`.

### DIFF
--- a/src/target/codegen_ascend_pto.cc
+++ b/src/target/codegen_ascend_pto.cc
@@ -437,6 +437,29 @@ CodeGenTileLangAscendPto::GetCompareMaskInfo(const CallNode *dst_call,
   int32_t col = shape[1].as<IntImmNode>()->value;
   int32_t valid_col = shape[3].as<IntImmNode>()->value;
 
+  // Memory planning may reuse a wider tensor buffer for the packed uint8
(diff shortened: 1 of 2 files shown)


z00520135 (approved):

ChaoyangJi (approved):
approve

PR #1755, opened 2026-09-08
[Bugfix] Resolve memory-planning kills at scope exits

### LINKED ISSUE #1765: [Bug] Memory planner can end liveness at a scope opening event
## Bug description

Ascend memory planning can associate a scope-exit kill with the opening event of the same scope. The resulting liveness interval ends too early, allowing a later allocation to overlap a buffer that is still read inside the scope and causing silent data corruption.

## Expected behavior

Resolve kill events to the closing occurrence so a buffer remains live through its final in-scope use.

## Regression coverage

Inspect generated UB allocations for overlap and verify the corresponding NPU result.

Proposed fix: #1755

wzzll123:
Closes #1765

## 变更摘要

Scope open/close entries share the same statement pointer in the memory planner linear sequence. `FindEventIndex` always selected the first occurrence, so a KILL moved to an enclosing scope could resolve to the opening entry, before GEN. The inverted live interval allowed still-live UB storage to be reused and silently corrupted results.

GEN now resolves to the first matching entry and KILL to the last matching entry.

## 变更类型

- [x] `[Bugfix]` Bug 修复

## 2. 框架及接口代码合入标准

- [x] Added a focused regression that verifies allocation offsets and NPU numerical correctness.
- [x] The change is limited to scope-event position resolution in Ascend memory planning.
- [x] No new operator; manifest and performance requirements are not applicable.
- [x] No public API or documentation change.

## 3. PR 提交规范

- [x] English `[Bugfix]` commit message.
- [x] PR contains only the fix and its regression.

## 其他补充

Original validation: the regression changed from overlapping `bigB/slabA` to safe reuse of the dead `bigC`; 72 memory-planning/copy/select language tests passed; conv2d local verification passed 80/80 cases.

### DIFF
--- a/src/transform/ascend_memory_planning.cc
+++ b/src/transform/ascend_memory_planning.cc
@@ -445,9 +445,28 @@ private:
         const auto &vec = use_gen ? event.gen : event.kill;
         auto it = std::find(vec.begin(), vec.end(), buffer);
         if (it != vec.end()) {
-          for (size_t i = 0; i < linear_seq_.size(); ++i) {
-            if (linear_seq_[i].stmt == event_pair.first) {
-              return static_cast<int64_t>(i);
+          // Scope statements (For/While/Let/If/Attr/Assert) appear twice in
+          // linear_seq_: once as the opening entry and once as the closing
+          // entry, sharing the same stmt object pointer.  A KILL attached to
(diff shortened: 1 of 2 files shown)


ChaoyangJi:
CI 没有通过，请检查问题。

PR #1756, opened 2026-09-08
[Bugfix] Use physical row widths for 2D subregion copies

### LINKED ISSUE #1767: [Bug] 2D subregion copies use sliced widths as physical row strides
## Bug description

GM-to-UB and UB-to-GM copies of a 2D subregion derive DMA row strides from the sliced region width. For a subregion embedded in a wider physical row, this advances to the wrong next-row address and corrupts data.

## Expected behavior

Use the physical source/destination row width when calculating inter-row strides while retaining the sliced width as the transferred row length.

## Regression coverage

Test both load and store directions with nonzero row/column offsets and padded physical rows.

Proposed fix: #1756

wzzll123:
Closes #1767

## 变更摘要

For a 2D copy whose on-chip endpoint is a subregion of a wider physical buffer, copy lowering used the region width as the template row width. The generated inter-row stride was therefore zero and rows after the first landed at incorrect addresses.

This PR uses the physical UB row width for both GM→UB destination stride and UB→GM source stride. Full-row and 1D copies are unchanged.

## 变更类型

- [x] `[Bugfix]` Bug 修复

## 2. 框架及接口代码合入标准

- [x] Added representative fp16/fp32 NPU tests for both copy directions.
- [x] Tests verify the copied rectangle bit-exactly and verify surrounding padding.
- [x] No new operator; manifest/performance requirements are not applicable.
- [x] No public API change.

## 3. PR 提交规范

- [x] English `[Bugfix]` commit message.
- [x] PR contains only the symmetric stride fixes and their shared regression.

## 其他补充

Original validation: the four 2D-subregion cases passed on NPU; left-pad, tail, misalignment and pad-value copy suites also passed (14 tests total).

### DIFF
--- a/src/op/ascend.cc
+++ b/src/op/ascend.cc
@@ -244,7 +244,20 @@ Stmt AscendCopy::Lower(const LowerArgs &T, arith::Analyzer *analyzer) const {
       // shape for a full copy, or the const slice width), so every existing
       // caller is unchanged -- only the previously-uncompilable runtime-slice
       // case changes.
+      //
+      // For a 2D+ destination the template column dim must be the buffer's
+      // PHYSICAL row width (dst->shape[last]), not the copy-region width
+      // (dst_extents[last]): the helper derives the dst inter-block stride as
+      // (dstN - maskShapeN) * sizeof(T) / 32, so a sub-region copy into a wider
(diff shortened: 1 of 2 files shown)


ChaoyangJi:
CI 没有通过，请检查问题。

PR #1761, opened 2026-09-08
[Bugfix] Flatten multidimensional scalar buffer indices

### LINKED ISSUE #1763: [Bug] Multidimensional scalar tile operands ignore trailing indices
## Bug description

When a multidimensional UB `BufferLoad` is used as the scalar operand of `T.tile.add`, `T.tile.mul`, and related operations, `binary_op()` passes only `indices[0]` to `GetValue`. For example, `n[i, j]` becomes `GetValue(i)` and silently ignores `j`.

## Expected behavior

Compute the scalar element offset from all indices, including explicit buffer strides and element offsets.

## Regression coverage

Use varying elements from a 2D UB buffer as scalar multipliers and compare AscendC/PTO results numerically.

Proposed fix: #1761

wzzll123:
Closes #1763

## 变更摘要

Fix scalar tile operands loaded from multidimensional buffers.

`binary_op()` previously passed only `indices[0]` to the scalar `GetValue` lowering. For a load such as `n_ub[i, j]`, this generated `GetValue(i)` and silently ignored `j`, producing incorrect numerical results on both AscendC and PTO.

The scalar offset now uses TVM Buffer's `offset_of(indices)`, which handles every dimension as well as explicit strides and element offsets.

## 变更类型

- [x] `[Bugfix]` Bug 修复
- [x] `[Testing]` 测试

## 测试

- [x] Added a 2D UB scalar-operand numerical regression for AscendC and PTO.
- [x] Ruff 0.16.6 format and lint pass for all changed Python files.
- [x] Python syntax compilation passes.
- [x] `git diff --check` passes.
- [x] Original NPU validation passed on both AscendC and PTO.

## 风险与边界

- One-dimensional loads retain the same effective offset.
- The fix reuses TVM's existing Buffer offset computation instead of duplicating row-major arithmetic.
- No public API changes.

## 提交前确认

- [x] Commit message is English and contains `[Bugfix]`.
- [x] The PR contains one focused fix and its regression.

### DIFF
--- a/tilelang/language/ascend_tile.py
+++ b/tilelang/language/ascend_tile.py
@@ -987,13 +987,16 @@ def binary_op(
     if isinstance(src1, BufferLoad):
         buffer_1 = src1.buffer
         indices_1 = src1.indices
+        # Account for every index, explicit strides, and the element offset.
+        # Using only the first index would lower n[i, j] to GetValue(i).
+        scalar_offset = buffer_1.offset_of(indices_1)[0]
         return T.call_intrin(
             "handle",
             tir.op.Op.get(f"tl.ascend_{op}s"),
             dst_ptr,
             src0_ptr,
             buffer_1.access_ptr("r"),
-            indices_1[0],
+            scalar_offset,
             size_0,
         )
 
--- a/testing/python/language/test_tilelang_ascend_language_scalar_from_2d_operand.py
+++ b/testing/python/language/test_tilelang_ascend_language_scalar_from_2d_operand.py
@@ -1,0 +1,63 @@
+# Copyright (c) Tile-AI Corporation.
+# Licensed under the MIT License.
+
+# Regression for using an element of a 2D UB tile as a scalar tile operand.
+
+import pytest
+import torch
+
+import tilelang
+import tilelang.language as T
+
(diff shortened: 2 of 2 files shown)


tilelang/language/ascend_tile.py
```diff
@@ -987,13 +987,16 @@ def binary_op(
     if isinstance(src1, BufferLoad):
         buffer_1 = src1.buffer
         indices_1 = src1.indices
+        # Account for every index, explicit strides, and the element offset.
+        # Using only the first index would lower n[i, j] to GetValue(i).
+        scalar_offset = buffer_1.offset_of(indices_1)[0]
```
ChaoyangJi, line 992: 同步修复select/compare的多维标量索引

本次只修复了binary_op,但select'的BufferLoad分支仍在约第809行传入indices_.1[0],compare分支仍在约第1893行传入'indices_1[0]。这两个API都支持BufferLoad标量；因此table[i,j]仍可能被降成只读取GetValue()`,在i>0时得到错误结果。建议抽出统一的flatten helper(如buffer..offset_of(indices),并为select/compare增加二维标量索引回归测试。

PR #1770, opened 2026-09-08
[Bugfix] Clear partial GM-to-L1 copies at ring-slot bases

### LINKED ISSUE #1769: [Bug] Partial GM-to-L1 copies leave stale data in reused ring slots
## Bug description

The AscendC `copy_gm_to_l1` codegen enables tail-padding clear only when the flat destination offset is zero. In a double-buffered or ring-buffered prefetch, a new slot starts at a nonzero tile-aligned offset, so a partial K-tail copy is incorrectly treated as a subregion update and is not cleared.

When the slot is reused, MMA consumes stale elements left by its previous tenant and produces incorrect convolution or GEMM results.

## Expected behavior

Treat both offset zero and a provably tile-aligned ring-slot base as primary copies that own tail-padding initialization. Keep genuine splice/subregion copies non-clearing so they do not overwrite data already assembled in the same tile.

## Regression coverage

Run a two-slot prefetching matmul with multiple K chunks and a partial final chunk, then compare the result against Torch.

Proposed fix will be submitted from `codex/fix-ascendc-l1-ring-clear`.

wzzll123:
Closes #1769

## 变更摘要

Fix stale L1 data when a partial GM-to-L1 tail copy targets a reused double-buffer/ring-buffer slot.

The previous `need_clear` heuristic only recognized flat destination offset zero as a primary copy. A ring-slot base is nonzero but aligned to the complete tile size, so tail padding was skipped and MMA could consume values left by the slot's previous tenant.

The codegen now also recognizes offsets that the TVM arithmetic analyzer can prove are multiples of the destination tile element count. Non-aligned splice/subregion updates continue to skip clearing.

## 变更类型

- [x] `[Bugfix]` Bug 修复
- [x] `[Testing]` 测试

## 测试

- [x] Added an AscendC two-slot prefetch regression with three K chunks and a partial final chunk.
- [x] Ruff 0.16.6 format and lint pass.
- [x] clang-format 18 passes.
- [x] Original NPU regression passed; the simplified Flash Attention set was also reverified.

## 风险与边界

- The additional clear is emitted only when tile alignment is statically provable.
- Non-aligned vertical-merge/splice copies preserve the existing non-clearing behavior.
- This PR targets the AscendC backend; it does not modify PTO.

### DIFF
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -2898,14 +2898,39 @@ void CodeGenTileLangAscend::CopyCodegen(const CallNode *op) {
     }
 
     // copy_gm_to_l1: append a `need_clear` flag so the helper only zero-inits
-    // the full L1 tile for the primary copy (dst offset 0). Sub-region copies
-    // (non-zero dst offset, e.g. the second DMA of a splice / vertical-merge
+    // the full L1 tile for the primary copy. Sub-region copies (dst offset not
+    // aligned to a whole tile, e.g. the second DMA of a splice / vertical-merge
     // pattern) must skip the clear, otherwise they clobber data already written
     // into the same NZ tile. dst_offset_expr is the scalar start offset of the
     // destination tile; when it is 0 this DMA targets the tile base and is the
-    // natural owner of the tail-padding clear.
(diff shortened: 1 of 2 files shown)


ChaoyangJi:
CI 没有通过，请检查问题。

PR #1772, opened 2026-09-08
[Feature] Use the hardware transpose path for BF16

### LINKED ISSUE #1771: [Feature] Use the hardware block-transpose path for BF16
## Problem

Large BF16 `T.tile.transpose` operations on the AscendC backend fall back to nested scalar `GetValue`/`SetValue` loops because CANN does not instantiate `TransDataTo5HDImpl<bfloat16_t>`. The operation is a bit-pattern movement, so the scalar fallback is unnecessary and extremely slow.

Measured on `[32, 1024, 1024]`, the fallback took about 24.2 ms versus 0.284 ms with the width-compatible hardware path.

## Requested behavior

Use the hardware block-transpose path for aligned BF16 tiles by instantiating the bitwise movement with a supported 16-bit payload type. Preserve the existing scalar fallback for unsupported shapes.

## Acceptance criteria

- BF16 output is bit-exact against `torch.permute`.
- Generated code contains the hardware `TransDataTo5HDImpl` path.
- Existing FP16/FP32 transpose behavior is unchanged.

Related to the broader transpose performance issue #1281.

wzzll123:
Closes #1771

Related to #1281

## 变更摘要

Move aligned BF16 tile transposes from the scalar fallback to the Ascend hardware block-transpose path.

`TransDataTo5HD` moves raw 16-bit payloads, but CANN does not provide a `bfloat16_t` template instantiation. `transpose_block` now uses `uint16_t` pointers and `TransDataTo5HDImpl<uint16_t>` for BF16 while retaining the original element types for FP16/FP32.

## 变更类型

- [x] `[Feature]` 新功能/后端能力
- [x] `[Testing]` 测试

## 测试

- [x] Added a pytest-discoverable AscendC BF16 transpose regression.
- [x] Output is checked bit-exactly against `torch.permute`.
- [x] Generated source is checked for the hardware block-transpose path.
- [x] Original transpose suite: 12 cases passed.
- [x] Original msprof result for `[32,1024,1024]`: about 24,167 us → 284 us (85×).

## 风险与边界

- Only BF16 payload typing changes; the operation remains bitwise.
- Shape eligibility for `transpose_block` is unchanged.
- This targets the AscendC backend and does not change PTO.

### DIFF
--- a/src/tl_templates/ascend/common.h
+++ b/src/tl_templates/ascend/common.h
@@ -1902,6 +1902,9 @@ Sum_experiment(const LocalTensor<T> &dst, const LocalTensor<T> &src,
 template <typename T, uint32_t H, uint32_t W>
 CATLASS_DEVICE void transpose_block(LocalTensor<T> const &dst,
                                     LocalTensor<T> const &src) {
+  // TransDataTo5HD moves bit patterns and CANN does not instantiate it for
+  // bfloat16_t. Reuse the 16-bit integer instantiation for BF16 payloads.
+  using MovT = std::conditional_t<std::is_same_v<T, bfloat16_t>, uint16_t, T>;
   constexpr uint32_t blockSize = 32 / sizeof(T);
   constexpr uint32_t highBlock = H / 16;
   constexpr uint32_t repeat = W / blockSize;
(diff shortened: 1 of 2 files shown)


ChaoyangJi:
CI 没有通过，请检查问题。

PR #1774, opened 2026-09-08
[Bugfix] Diagnose misaligned on-chip copy addresses

### LINKED ISSUE #1773: [Bug] Misaligned UB/L1 copy addresses are not diagnosed at compile time
## Bug description

AscendC copy lowering can emit a `DataCopy`/`DataCopyPad` whose UB or L1 `LocalTensor` starts at a compile-time constant address that is not 32-byte aligned. CANN requires the LocalTensor start address to be 32-byte aligned, while the GM side only requires byte alignment.

The compiler currently lets this invalid program reach BiSheng or hardware, where the resulting failure is difficult to associate with the source-level copy.

## Expected behavior

When an on-chip copy offset is a compile-time constant, reject a non-32-byte-aligned UB/L1 base during code generation and report the operation, buffer, element offset, byte offset, and required alignment. Dynamic offsets should remain accepted because their runtime value may be aligned.

## Regression coverage

- Reject a constant FP32 UB offset of one element (4 bytes).
- Accept and execute a constant FP32 UB offset of eight elements (32 bytes).

This issue covers the AscendC backend only.

wzzll123:
Closes #1773

## 变更摘要

Diagnose statically misaligned UB/L1 copy addresses in AscendC code generation.

CANN requires the LocalTensor side of `DataCopy`/`DataCopyPad` to start at a 32-byte-aligned address. This PR checks constant on-chip offsets for GM→UB, UB→GM, UB→UB, and GM→L1 copies and reports a source-oriented compile error before an invalid operation reaches BiSheng or hardware.

Runtime-dependent offsets remain accepted because they may be aligned at execution time. GM addresses are not restricted by this check.

## 变更类型

- [x] `[Bugfix]` Bug 修复
- [x] `[Testing]` 测试

## 测试

- [x] Constant FP32 UB offset 1 (4 bytes) is rejected with an alignment diagnostic.
- [x] Constant FP32 UB offset 8 (32 bytes) compiles and executes correctly.

## 风险与边界

- Only compile-time constant offsets are rejected.
- Unknown template dtypes skip the check to avoid false positives.
- This modifies AscendC codegen only; PTO is unchanged.

### DIFF
--- a/src/target/codegen_ascend.cc
+++ b/src/target/codegen_ascend.cc
@@ -14,6 +14,8 @@
 
 #include <tvm/tir/expr.h>
 
+#include <algorithm>
+#include <cctype>
 #include <cmath>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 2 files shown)


src/target/codegen_ascend.cc
```diff
@@ -2889,6 +2891,70 @@ void CodeGenTileLangAscend::CopyCodegen(const CallNode *op) {
...
+    };
+    int elem_bytes = tmpl_dtype_bytes(op_name);
+    if (elem_bytes > 0) {
+      auto check_side_align = [&](const PrimExpr &off_expr,
+                                  const std::string &buf_name,
+                                  const char *side) {
+        const auto *imm = off_expr.as<IntImmNode>();
+        if (imm == nullptr)
+          return; // runtime offset -> let through
+        int64_t byte_off = imm->value * elem_bytes;
```
ChaoyangJi, line 2933: UB-UB source alignment uses destination dtype width

copy_ub_to_ub的模板参数顺序是<dst_dtype, src_dtype, len>,而这里的elem_bytes只解析第一个类型，并同时用于src_offset_expr与dst_offset_.expr。
混合dtype时会误判：例如copy_ub_to_ub<float,,half,>中源偏移8是16B,未满足32B对齐，但当前按float的4B计算成32B而放过；反向组合还可能把合法源偏移误拒。

请分别按源/目的Buffer dtype计算字节偏移，并补充mixed-dtype UB-UB的源、目的两侧测试。

wzzll123:
Addressed in c56c7de0. `copy_ub_to_ub<dst_dtype, src_dtype, len>` now parses and checks destination and source offsets using their respective element widths. Added mixed-dtype regressions for a misaligned half source, a misaligned float destination, and the both-32B-aligned control case. The full alignment regression file passed in an isolated Docker/CANN build.

PR #1776, opened 2026-09-08
[Feature] Add a coalesced GM-to-L1 row-copy path

### LINKED ISSUE #1775: [Performance] Coalesce contiguous 32-byte-row GM-to-L1 copies
## Performance problem

GM-to-L1 copies of an NC1HWC0-style band with 16 two-byte elements per row are currently lowered through the generic TLA path, which issues one 32-byte transfer per row. This leaves the MTE2 pipeline dominant in convolution workloads even though the GM rows are physically contiguous.

The linear and zN byte layouts are identical for a complete 32-byte row of 16-bit elements, so consecutive rows can be transferred as one block without a layout conversion.

## Proposed optimization

Use one coalesced `DataCopy` when all of the following are true:

- source and destination inner widths are complete and equal;
- there is no column tail;
- the row is exactly 32 bytes;
- the element width is 16 bits.

Retain the generic TLA path for FP32, column tails, strided rows, and other layouts.

## Validation

- Numerical GM → L1 → L0 → identity MMA regression.
- Original copy microbenchmark: about 1128 us → 313 us.
- Original Conv2D case: about 8767 us → 1339 us.
- Original Conv2D verification: 19/19 local and 20/20 cann-bench cases passed.

wzzll123:
Closes #1775

## 变更摘要

Add a coalesced AscendC GM-to-L1 path for contiguous bands whose complete row is exactly one 32-byte block of 16-bit elements.

For this specific layout, linear and zN storage are byte-identical. The new path transfers all valid rows with one `DataCopy` instead of issuing a separate 32-byte TLA transfer for every row.

The guard requires equal complete inner widths, no column tail, a 32-byte row, and 16-bit elements. FP32 and all non-equivalent layouts remain on the generic path.

## 变更类型

- [x] `[Feature]` 后端性能路径
- [x] `[Testing]` 测试

## 测试与性能

- [x] Added an independent FP16 GM → L1 → L0 → identity MMA numerical regression.
- [x] Original copy microbenchmark: about 1128 us → 313 us.
- [x] Original Conv2D case: about 8767 us → 1339 us (6.5×).
- [x] Original validation: Conv2D 19/19 local, 20/20 cann-bench, transpose 12/12, and Mamba2Bwd 50/50.

## 风险与边界

- A column tail explicitly disables the fast path to prevent an out-of-range full-row read.
- FP32 C0=8 remains generic because its linear and fractal layouts are not byte-identical.
- PTO is unchanged.

### DIFF
--- a/src/tl_templates/ascend/common.h
+++ b/src/tl_templates/ascend/common.h
@@ -71,6 +71,20 @@ copy_gm_to_l1(LocalTensor<T> dstTensor, GlobalTensor<T> srcTensor,
         {1, static_cast<uint16_t>(dstM * dstN * sizeof(T) / 32), 0, 0});
     AscendC::PipeBarrier<PIPE_MTE2>();
   }
+  // A full-width 32-byte row of 16-bit elements has byte-identical linear and
+  // zN representations. Coalesce consecutive rows into one DMA instead of
+  // issuing one 32-byte transfer per row. FP32 C0=8 does not have this layout
+  // equivalence and intentionally stays on the generic path. A column tail
+  // must also stay generic to avoid reading a complete row past valid GM data.
+  if (realSrcN == dstN && tailN == dstN && dstN * sizeof(T) == 32 &&
+      sizeof(T) == 2) {
+    AscendC::DataCopy(dstTensor, srcTensor,
+                      AscendC::DataCopyParams(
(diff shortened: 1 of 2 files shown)


ChaoyangJi:
CI 没有通过，请检查问题。

